PR #23, opened 2018-06-07
Rocko32

htot:
I force pushed rocko64-acpi (sorry). 

This branch is then on top of rocko64-acpi, but switches to 32b and turns off acpi. 

Nevertheless we can not assume bios has no-acpi so we turn acpi=off when booting from eemc. When booting from usb/sd card you still need to do that manual I use `setenv acpi 'acpi=off quiet'` prior to `run usbboot`. Both branches provide a new u-boot, so if you reflash that with this one you will have no-acpi version.

I also added a temporary patch to enable i2c-6. This will only work on the no-acpi kernel. The bus detects, but I haven't tested connecting a device. Maybe you can, mraa should work normal.

Running make image now (so partial rebuild, ~~so far~~ no problems except the CPU fan running at 3000rpm, is a bit hot here)


andy-shev:
Good question. Since I'm using UART as a console only and not much relied on Bluetooth functionality, I can't stand for 32-bit only case. Though, for DIY people UART iisue might be crucial and thus better to stick with reliable solution.

alext-mkrs:
Thanks. And what do you think from the general kernel support prospective - I e.g. read somewhere that 32 bit x86 gets much less love these days as x86-64 is prevalent (e.g. those Meltdown/Spectre patches were planned for later)? I haven't yet spent any significant time verifying that, so I'm curious about your opinion here.

htot:
The patches have either gone into upstream (so won't apply) or experimental and accidentally committed. The cfg's are nw unneeded as Andy's kernel has them.

alext-mkrs:
Got it, thanks.

alext-mkrs:
So I'm getting a kernel build failure, which looks like some architecture mismatch:
```
<...>
|   WRAP    arch/x86/include/generated/asm/mm-arch-hooks.h
|   CC       /edison_dev/yocto/vanilla32/out/linux64/build/tmp/work/edison-poky-linux/linux-yocto/4.16.0-r0/linux-edison-standard-build/tools/objtool/orc_gen.o
|   CC      scripts/mod/empty.o
| cc1: error: code model 'kernel' not supported in the 32 bit mode
| cc1: sorry, unimplemented: 64-bit mode not compiled in
| /edison_dev/yocto/vanilla32/out/linux64/build/tmp/work-shared/edison/kernel-source/scripts/Makefile.build:324: recipe for target 'scripts/mod/empty.o' failed
| make[4]: *** [scripts/mod/empty.o] Error 1
| /edison_dev/yocto/vanilla32/out/linux64/build/tmp/work-shared/edison/kernel-source/scripts/Makefile.build:583: recipe for target 'scripts/mod' failed
| make[3]: *** [scripts/mod] Error 2
| /edison_dev/yocto/vanilla32/out/linux64/build/tmp/work-shared/edison/kernel-source/Makefile:566: recipe for target 'scripts' failed
| make[2]: *** [scripts] Error 2
<...>
```

That's on Ubuntu 16.04 x86-64, which built official (32-bit) + our latest 64-bit images just fine. Investigating...

htot:
Hmm. I have that when I build with `bitbake -k edion-image` after switching from 64b to 32b. If you are using that, you could try `bitbake linux-yocto -c cleansstate` and then `bitbake -k edion-image` again.

I've been building the kernel a lot lately and got sick of doing that and just do `make image`. It's a bit slower as it deletes more (u-boot), but at least it stops complaining.

htot:
I can reproduce. First I did `make clean` followed by `make image` which gave the error. Then tried `make cleansstate` then `make image`. Which worked.
But if you are right it worked because it was the 2nd attempt.

alext-mkrs:
Right, that's similar to what I observe

htot:
Build now completed. Next building sdk.

alext-mkrs:
So looks like the first time `linux-yocto` build runs, it creates the kernel config with `CONFIG_X86_64=y`, for whatever reason. However using `clean`/`cleansstate` and manually removing `.config` and then runnning `do_configure` task does create a proper config. I've checked all our settings in kernel-related recipes and everything looks ok in general. I'm now trying `cleanall` (i.e. it will redownload the source) to see if it changes anything, but this is quite strange (but at least understood now).

htot:
Very strange. Maybe we can find from the logs in which order the recipes ran, and from there which created the wrong config?

alext-mkrs:
Yeah, I'll take a look - but I'd be quite surprised if task order would change, that's pretty fixed. I could think of some variables being set with delays - but then again I don't yet understand how to reproduce at will, the clean/cleansstate should've done that.

I run bitbake manually, but those makefile targets are essentially the same, just with checks if it's already setup and so on - additional "helpers" I don't really need.

That git fetch of the kernel is really slow, still at 43% at the moment, so looks like this will continue tomorrow.

alext-mkrs:
Oh, sorry - you meant *recipes* not *tasks* - generally other recipes cannot influence that. That's surely something happening within the `linux-yocto` recipe and looks like there's some difference between "the first time" (and the definition of that is what I'm trying to understand right now) and the "other ones".

htot:
No, I did mean the task order of linux-yocto recipe, just phrased it wrong.

alext-mkrs:
Okay, so `cleanall` and redownloading didn't help, as I suspected. So let me try one last thing before I ditch that for now and start looking into regulatory DB stuff, which should be more fun. `make clean` you said reproduces that - and it simply yanks the `out` dir as a whole, let me try that.

BTW, what exactly was the problem with HSU you faced on 64bit? Looking through all the public announcements about Merrifield, it's advertised as 64bit core/SoC, so it doesn't look like it would be a hack to run it like that. So maybe we could just fix the stuff to work on 64 instead of resorting back to 32, which reportedly gets less love from kernel community and produces  bitbake glitches :D

htot:
I was on 64b, then checked out 32b, followed by `make clean`. But that only cleans `out`, not `sstate-cache`. So I thought that to be the reason why it tries to build 64b kernel again. `make cleansstate` seems to fix that. Though in the past I just deleted the contents of `bbcache/sstate-cache`/ (but kept `downloads`, so not having to go through that again).

Just checked my `log.task_order` for linux-yocto, but unfortunately it starts with `do_cleansstate`. Interesting stuff would be before that.

alext-mkrs:
I've checked the task order - it's fine. First kernel_configme, then configure, as expected. And the result is the same - proper config. So looks like it's something strictly related to the very first time it builds after setup. The only other thing I've noticed is that the NATIVELSBSTRING value changes after it downloads that uninative thing. Maybe that's related, not sure. I'll probably look into regulatorydb stuff now and the get back to this.

htot:
Alright. I think we need to get hostapd working first. As I understand the regdb was only needed for a 5G AP. I tried it last weeked, the postbuild script wants to manipulate the hostapd conf file on first run, but the location has changed. Without modifying anything I see the AP on my phone with an open connection, but connecting didn't work. With AP working we could try to reinstate the factory config script / web server... never got to that point.

alext-mkrs:
All right - I've got the 32bit image ready, just need to get the dust off my Edison, flash it and look into hostapd.

As a note to self to document findings: while building the above I've poked around that kernel build issue a bit more and looks like the reason is that for whatever reason when being run for the very first time after setup, the `do_kernel_configme()` doesn't produce a proper `.config`, like if the `merge_config.sh` didn't work properly, while the logs (including the one from that script) seemingly look ok. Need to look further from there.

htot:
Hmm. Something wrong with my config fragments? Had that before, but how would that change 32b into 64b.

You might appreciate this:
```
root@edison:~# modprobe -i i2c-dev
root@edison:~# i2cdetect -r 1                                                                                                                                                                                                                                                  
WARNING! This program can confuse your I2C bus, cause data loss and worse!                                                                                                                                                                                                     
I will probe file /dev/i2c-1 using read byte commands.                                                                                                                                                                                                                         
I will probe address range 0x03-0x77.                                                                                                                                                                                                                                          
Continue? [Y/n]                                                                                                                                                                                                                                                                
     0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f                                                                                                                                                                                                                            
00:          -- -- -- -- -- -- -- -- -- -- -- -- --                                                                                                                                                                                                                            
10: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --                                                                                                                                                                                                                            
20: UU UU UU UU -- -- -- -- -- -- -- -- -- -- -- --                                                                                                                                                                                                                            
30: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --                                                                                                                                                                                                                            
40: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --                                                                                                                                                                                                                            
50: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --                                                                                                                                                                                                                            
60: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- --                                                                                                                                                                                                                            
70: -- -- -- -- -- -- -- --                                                                                                                                                                                                                                                    
root@edison:~# i2cdetect -r 6                                                                                                                                                                                                                                                  
WARNING! This program can confuse your I2C bus, cause data loss and worse!                                                                                                                                                                                                     
I will probe file /dev/i2c-6 using read byte commands.
I will probe address range 0x03-0x77.
Continue? [Y/n] 
     0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f
00:          -- -- -- -- -- -- -- -- -- -- -- -- -- 
10: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
20: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
30: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
40: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
50: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
60: -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- -- 
70: -- -- -- -- -- -- -- --  
```                       
If you have some i2c based shield lying around, you might want to plug it :-)

alext-mkrs:
Yeah, that's nice :)

meta-intel-edison-bsp/recipes-kernel/linux/linux-yocto_4.16.0.bb
```diff
@@ -16,6 +16,10 @@ SRC_URI = "git://github.com/htot/linux.git;protocol=https;branch=eds-4.16.0-unif
         file://configfs.cfg \
         file://usb_dwc3.cfg \
         "
+SRC_URI_append = " ${@bb.utils.contains('DISTRO_FEATURES', 'acpi', '', 'file://file://i2c_modules.cfg', d)}"
```
htot: I added this to make i2c modules only for no-acpi version.

meta-intel-edison-bsp/recipes-bsp/u-boot/u-boot-common_2018.05.inc
```diff
@@ -25,17 +25,17 @@ EDISON-PATCHES = "\
 "
 
 SRC_URI = "${@bb.utils.contains('DISTRO_FEATURES', \
-	'acpi', 'git://github.com/htot/u-boot.git;branch=acpi-v2018.03;protocol=https', \
+	'acpi', 'git://github.com/htot/u-boot.git;branch=acpi-v2018.05;protocol=https', \
```
htot: Andy released his new 2018-05. I tested by building and installing the version with acpi, using the rocko64-acpi branch. I did not yet test the no-acpi version (rocko32 branch)

htot:
Changed i2c and added u-boot. I put my review comments in the commits. Sorry for rebasing.

alext-mkrs:
Okay, I was going to work on hostapd tomorrow, so won't rebuild right away, but will test afterwards.

alext-mkrs:
So the hostapd stuff aside - let me try building the updated version and let's merge it (if it's ok) - I believe we better chunk the changes instead of keeping this long PR open and updating it continuously.

meta-intel-edison-bsp/recipes-kernel/linux/linux-yocto_4.16.0.bb
```diff
@@ -5,26 +5,31 @@ KBRANCH ?= "standard/base"
...
         file://smsc95xx.cfg \
         file://bt_more.cfg \
         file://i2c_chardev.cfg \
         file://configfs.cfg \
+        file://usb_dwc3.cfg \
         "
+SRC_URI_append = " ${@bb.utils.contains('DISTRO_FEATURES', 'acpi', '', 'file://file://i2c_modules.cfg', d)}"
+
+# kernel patches
+SRC_URI_append = "${@bb.utils.contains('DISTRO_FEATURES', 'acpi', '', 'file://0001-Add-kernel-parameter-to-enable-i2c-6-pinctrl-mapping.patch', d)}"
```
alext-mkrs: This one also fails the build - needs a space right after the opening quotation mark, otherwise the content appended is being concatenated with the previous one.

meta-intel-edison-bsp/recipes-bsp/u-boot/u-boot-common_2018.05.inc
```diff
@@ -20,26 +20,22 @@ PE = "1"
 # otherwise we use upstream with the following patches
 
 EDISON-PATCHES = "\
-	file://0001-x86-tangier-Make-_CRS-for-BTH0-Serialized-to-avoid-w.patch \
-	file://0002-x86-Fix-reference-to-QEMU-variant-of-write_acpi_tabl.patch \
-	file://0003-x86-zImage-Move-subarch-assignment-out-of-cmd_line-c.patch \
-	file://0004-x86-zImage-Propagate-acpi_rsdp_addr-to-kernel-via-bo.patch \
 	file://0005-HACK-dfu-add-optional-timeout.patch \
 	file://0006-HACK-mmc-sdhci-SDHCI-controllers-also-need-power.patch \
```
alext-mkrs: This one no longer applies cleanly due to context changes in `edison.c`. I'll prepare a patch shortly.

meta-intel-edison-bsp/recipes-kernel/linux/linux-yocto_4.16.0.bb
```diff
@@ -5,26 +5,31 @@ KBRANCH ?= "standard/base"
...
-        file://usb_dwc3.cfg \
+SRC_URI = "git://github.com/htot/linux.git;protocol=https;branch=eds-4.16.0-unified \
         file://ftdi_sio.cfg \
         file://smsc95xx.cfg \
         file://bt_more.cfg \
         file://i2c_chardev.cfg \
         file://configfs.cfg \
+        file://usb_dwc3.cfg \
         "
+SRC_URI_append = " ${@bb.utils.contains('DISTRO_FEATURES', 'acpi', '', 'file://file://i2c_modules.cfg', d)}"
```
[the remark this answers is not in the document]
alext-mkrs: Fixed in my commit

meta-intel-edison-bsp/recipes-kernel/linux/linux-yocto_4.16.0.bb
alext-mkrs: Fixed in my commit

meta-intel-edison-bsp/recipes-bsp/u-boot/u-boot-common_2018.05.inc
alext-mkrs: Fixed in my commit

alext-mkrs:
All right, so I've pushed one commit into the PR with fixes for my comments (all but one - that attribution thing is worth to be checked). Please just squash it when merging.

Now the PR is build- and boot-tested, looks ok. Let's get this merged and start stabilizing it, I see some things to check around wifi and all those official build remnants (I've created a couple of GH issues to track that).

meta-intel-edison-bsp/recipes-kernel/linux/files/0001-serial-8250_dma-use-linear-buffer-for-transmit.patch
```diff
@@ -0,0 +1,88 @@
+From 1a737b2e31e996dea02ab9a0823f0a81284d5d2e Mon Sep 17 00:00:00 2001
+From: Ferry Toth <ftoth@exalondelft.nl>
+Date: Sat, 17 Mar 2018 13:54:18 +0100
+Subject: [PATCH] serial: 8250_dma: use linear buffer for transmit
```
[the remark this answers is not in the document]
htot: Yeah, I did this. It's the meager result of months of work trying to get serial port working at 2Mb/s. 

On 32 bits this is enough to send messages in one piece without interchar gap.

On 64b we have a long interrupt latency causing the fifo to overrun. When I reduce the fifo threshold the chance of loosing bytes is greatly reduced, but between 1M - 4M messages the HSU locks up. This does not happen on 32b

htot:
Yes, see above. I wrote this. Andy didn't like it much I think and didn't want to take it. He doesn't like that I copy the circular buffer into a new linear buffer. To make it into the kernel I would probably need to revise this and submit 2 descriptors to the dmaengine. As I wasn't sure those would be handled consecutively by the dma handle (without interchargap) and the current patch works I didn't try this yet.

htot:
BTW I split up your commit into i2c: and u-boot: and squashed them with the both erratic commits. I am making rocko32 now to check that I didn't introduce new errors. 

Can you please check if I attributed your work correctly? I'm not sure what is the correct way or how to do that and would like to learn that. 

If it is ok as is, I'll pull rocko32 into master.

meta-intel-edison-bsp/recipes-kernel/linux/files/0001-serial-8250_dma-use-linear-buffer-for-transmit.patch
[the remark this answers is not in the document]
alext-mkrs: Ok, I see, so that's that thing related to both 32 vs 64 bits and interrupt latency. Probably worth looking into later on.

alext-mkrs:
The branch looks (and builds) good to me.

As for the attribution - with such small ones (and within the PR context where it merely fixed the typos right away instead of me commenting and you fixing), I'm not sure it's even worth it :) But typically it's done in a form of just a reference "based on original patch by XYZ", "idea inspired by XYZ" or somesuch. or preserving the commit author in git.

The Signed-off-by line typically means only that the signing off person confirms the patch was developed in accordance with the given project's policies and that the licence and copyright are appropriate. And duplicating it (with the same name) usually is not advised. Such a layered one is typically seen when there's an original author of the patch (signoff 1), which is picked to the repo by the maintainer, which confirms the patch validity and acceptance (signoff 2).

All in all in this specific case I'd say you could've just used my fixes without any specific attribution. And if there were full patches/commits to merge, then you could either (1) commit by yourself and attribute in the commit message or (2) just take the commit and merge/cherry-pick it, preserving the author (and you would be the committer, git separates these two).

alext-mkrs:
No worries, that's not necessary - but thanks!

PR #26, opened 2018-06-23
Fix repo references after moving to edison-fw org

alext-mkrs:
I've not corrected all of the references in `README.md` on purpose - as those currently left are about `meta-intel-iot-middleware` layer, which we don't use any more anyway and it's just that the whole readme needs an overhaul. I'll send a PR proposal later.

Other than that there were just a few fixes. The build is still running for me, so @htot, please don't merge it just yet - for now please take a look and tell me if it looks ok for you. When my build is [successfully] done, I'll merge it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,4 +27,4 @@ Currently I am tracking origin/master but I have created four additional branche
 
 # How to use this
 
-Yocto Morty / Pyro will build on Ubuntu Artful (17.10). For detailed instructions see the wiki [https://github.com/htot/meta-intel-edison/wiki](https://github.com/htot/meta-intel-edison/wiki)
+Yocto Morty / Pyro will build on Ubuntu Artful (17.10). For detailed instructions see the wiki [https://github.com/edison-fw/meta-intel-edison/wiki](https://github.com/edison-fw/meta-intel-edison/wiki)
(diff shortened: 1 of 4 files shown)


htot:
Looks fine too me

htot:
So how would the process work now? Let's say I have a commit for meta-acpi (which I have), I need to clone edison-fw/meta-acpi to htot/meta-acpi and push it there, right? Because right now (without cloning) pushing to htot/meta-acpi would go directly to edison-fw/meta-acpi (I think).

alext-mkrs:
So there are basically two options - you either clone the edison-fw/meta-acpi, make a "feature", temporary, branch and then send a pull request from that branch to master, or create a fork in htot/meta-acpi, create a branch and then send a pull request from that to edison-fw/meta-acpi:master.

The latter used to be the recommended GH flow. The difference is that you have a full total copy of the repo you can mess with, not influencing the main repo in any way.

Here's some description of the flow: https://help.github.com/articles/working-with-forks

Either way, I'd keep my previous proposal of having all changes through pull requests, so that everything is peer-reviewed, I've learned over time that even in the most simple-looking changes there are sometimes pitfalls one doesn't think of. And that helps keep everyone on the same page. They can be self-merged, but the review is IMHO quite important.

All right, my build test also went fine, so I'm merging this one - thanks for taking a look.

alext-mkrs:
This whole chapter may also be handy actually: https://help.github.com/categories/collaborating-with-issues-and-pull-requests/

PR #29, opened 2018-07-02
Sumo64 acpi

htot:
Don't merge this now. This patch set upgrades to sumo and bluntly disables stuff that causes build errors (except the first 3 - of which the first 2 should not go into master at all).  Unfortunately things you have fixed before (SDKTAROPTS patch and mraa bindings) broke again. A new thing is python-importlib.

There is a ton of warnings that we need to set a compatibility variable in our layers, but that doesn't hurt for now.

I seems sumo is solving the modules problem I had, but to be sure I need `make clean` and start again - tomorrow.


htot:
I see. Sumo applies the 0001-Add-Node-7.x-aka-V8-5.2-support.patch to swig, so we should be building nodejs 7+ (nodejs 8.9.4 provided by sumo should be working and we could drop the meta-nodejs layer). I am rebuilding from scratch now, and will try this when done.

htot:
Just verified, this patch set builds without error in one pass. Now building with nodejs 8.9.4 with mraa bindings.

alext-mkrs:
> Amazing, it just builds

Yeah, that's what I was hoping for :) Let me give it a try to see if my (different than yours) setup looks good as well.

htot:
SDK builds fine too and functions. I still need to test the rootfs.

alext-mkrs:
Builds fine for me too.

htot:
Initramfs installs (from manifest):
grub core2-64 2.02-r0
grub-common core2-64 2.02-r0
grub-editenv core2-64 2.02-r0
adding like 14Mb. Now we get error bzImage too big 19.6Mb, max 15.7Mb. Also, there is a new init script which overrules ours.

Need to do a little work here.

alext-mkrs:
I've checked this a bit and grub is being pulled in by `initramfs-module-install` recipe, which `RDEPENDS` on it. That module was added in poky commit 91972d58c9fc40cc1d7e568d2457936257f79b45 "initramfs-framework: add install module". I guess the same one may redefine the script (or that the bbappend we have, for initramfs-live-boot` is not included at all). Ran out of time today to see why this is happening and what are those recipes all about in a big picture, but hopefully this helps.

alext-mkrs:
So to get rid of the grub, the straightforward solution is below. Now, it needs to be analyzed why there were exclusions for `_live_install` but no `_module_install` - but given that commit I referenced above, there probably just hasn't been one, so no exclude was needed, so the below should probably be good enough. So only the script override left to tackle, I guess.

```patch
diff --git a/meta-intel-edison-distro/recipes-core/images/core-image-minimal-initramfs.bbappend b/meta-intel-edison-distro/recipes-core/images/core-image-minimal-initramfs.bbappend
index a014bf3..fa429db 100644
--- a/meta-intel-edison-distro/recipes-core/images/core-image-minimal-initramfs.bbappend
+++ b/meta-intel-edison-distro/recipes-core/images/core-image-minimal-initramfs.bbappend
@@ -9,10 +9,12 @@ PACKAGE_INSTALL += " kernel-module-mmc-block"
 PACKAGE_INSTALL += " kernel-module-sdhci-acpi"
 PACKAGE_INSTALL += " kernel-module-sdhci-pci"
 PACKAGE_INSTALL += " ${@bb.utils.contains('DISTRO_FEATURES', 'acpi', 'acpi-tables', '', d)}"
-PACKAGE_INSTALL_remove = " initramfs-live-install"
-PACKAGE_INSTALL_remove = " initramfs-live-install-efi"
+PACKAGE_INSTALL_remove = " initramfs-live-install initramfs-module-install"
+PACKAGE_INSTALL_remove = " initramfs-live-install-efi initramfs-module-install-efi"
```

htot:
Thanks. In the past I put these removes because of the same problem, cpio exceeding max. size.I guess I just looked at the recipe's installs and removed what was unneeded. And got the tip to clobber the kernel image from the mailing list.  I'll try this and fix the script, and while there see if there is anything to be learned from the new default script.

htot:
For some reason we are no longer pulling initramfs-live-boot but instead initramfs-framework. I think that explains the modules install as well as the wrong initscript. Instead of figuring out why, just remove unnecessary stuff and bbappend initramfs-framework.

With these last 2 patches I am able to boot Edison from usb. Will test booting from sd card, wifi, bluetooth tomorrow.

alext-mkrs:
Ok, found it:

```
commit 53078a00ceab5d6aaa61c77fcb73f0d4aa788e00
Author: California Sullivan <california.l.sullivan@intel.com>
Date:   Fri Apr 13 13:36:40 2018 -0700

    core-image-minimal-initramfs: use initramfs-framework for initialization

    initramfs-framework is more modular and expandable. This change was
    proposed in commit 28fc6ba761ed4a47efa7c43e7f7dff5e2fe72b5e
    "core-image-minimal-initramfs: use initramfs-framework by default" but
    reverted due to the selftests runqemu.RunqemuTests.test_boot_machine_iso
    and runqemu.RunqemuTests.test_boot_deploy_hddimg failing. Since then,
    the kinks have been worked out, and missing functionality that had been
    missed (non-EFI installation module) has been added.

    Since the PACKAGE_INSTALL variable was getting so long with all these
    individual modules getting added, I also introduced a new
    INITRAMFS_SCRIPTS variable to the core-image-minimal-initramfs recipe.
    This variable makes the recipe look much cleaner, and also allows easier
    replacement or additions to the scripts.

    Fixes [YOCTO #10987].

    (From OE-Core rev: 882ae0dcce2d96a7c286fc23b22b07972d3d8f93)

    Signed-off-by: California Sullivan <california.l.sullivan@intel.com>
    Signed-off-by: Ross Burton <ross.burton@intel.com>
    Signed-off-by: Richard Purdie <richard.purdie@linuxfoundation.org>
```

So now `/out/linux64/poky/meta/recipes-core/images/core-image-minimal-initramfs.bb` has this

```bash
INITRAMFS_SCRIPTS ?= "\
                      initramfs-framework-base \
                      initramfs-module-setup-live \
                      initramfs-module-udev \
                      initramfs-module-install \
                      initramfs-module-install-efi \
                     "

PACKAGE_INSTALL = "${INITRAMFS_SCRIPTS} ${VIRTUAL-RUNTIME_base-utils} udev base-passwd ${ROOTFS_BOOTSTRAP_INSTALL}"
```

and we can either add the `live-boot` back, or as you just did override `init` in `initramfs-framework`, both will be ok.

In addition, after analyzing what those recipes in `INITRAMFS_SCRIPTS` do, I velieve we can further remove `initramfs-module-setup-live` (seems to only needed if we use "rootfs as file").

meta-intel-edison-distro/recipes-core/initrdscripts/initramfs-framework_1.0.bbappend
```diff
@@ -1,6 +1,6 @@
 SUMMARY = "Live image init script"
 #LIC_FILES_CHKSUM = "file://${COREBASE}/meta/COPYING.MIT;md5=3da9cfbcb788c80a0384361b4de20420"
 
-FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
-SRC_URI += "file://init-live.sh \
+FILESEXTRAPATHS_prepend := "${THISDIR}/initramfs-framework:"
+SRC_URI += "file://init \
```
alext-mkrs: This one isn;'t really necessary - the main recipe has this already, so it's enough to just prepend our directory into the `FILESEXTRAPATHS` and it will get picked up.

meta-intel-edison-distro/recipes-core/images/core-image-minimal-initramfs.bbappend
```diff
@@ -9,8 +9,8 @@ PACKAGE_INSTALL += " kernel-module-mmc-block"
 PACKAGE_INSTALL += " kernel-module-sdhci-acpi"
 PACKAGE_INSTALL += " kernel-module-sdhci-pci"
 PACKAGE_INSTALL += " ${@bb.utils.contains('DISTRO_FEATURES', 'acpi', 'acpi-tables', '', d)}"
-PACKAGE_INSTALL_remove = " initramfs-live-install"
-PACKAGE_INSTALL_remove = " initramfs-live-install-efi"
+PACKAGE_INSTALL_remove = " initramfs-live-install initramfs-module-install"
```
alext-mkrs: Let's add `initramfs-module-setup-live` as well, unless you see a reason to preserve it.

meta-intel-edison-distro/recipes-core/images/edison-image.bb
```diff
@@ -81,7 +81,7 @@ IMAGE_INSTALL_append = " alsa-utils-alsamixer alsa-utils-alsactl alsa-utils-apla
 # removed python-imaging python-pygobject python-dbus 
 IMAGE_INSTALL_append = " python"
 IMAGE_INSTALL_append = " python-argparse"
-IMAGE_INSTALL_append = " python-distutils python-pkgutil python-audio python-image python-email python-netserver python-xmlrpc python-ctypes python-html python-json python-compile python-misc python-numbers python-unittest python-pydoc python-importlib"
+IMAGE_INSTALL_append = " python-distutils python-pkgutil python-audio python-image python-email python-netserver python-xmlrpc python-ctypes python-html python-json python-compile python-misc python-numbers python-unittest python-pydoc"
```
htot: I disabled importlib here, but don't know what the effect is. Also in the past removed certain python libs, that now could possibly be added back (pygobject?). I don't know python, so I have no idea what would be a sane set of libs. 

What do you think? Leave it and change when needed or try to stay as close as possible to the factory image?

lybtongji: @htot here (https://lists.yoctoproject.org/pipermail/yocto/2018-May/041101.html) said "python-importlib is now part of python*-core"

alext-mkrs: ...so this specific one is clear and in general I think it doesn't really matter that much - if anyone uses a library we accidentally drop, they'll let us know. And pip can always be used to install most of them anyway after the fact, so IMHO that's not a big deal.

alext-mkrs:
@htot, JFYI in the meanwhile I've started cleaning up layers on my branch - removed intel-iot-middleware, nodejs, added `LAYERSERIES_COMPAT`, so as soon as this one is in, I'll create a PR for that.

meta-intel-edison-distro/conf/distro/poky-edison.conf
```diff
@@ -6,17 +6,17 @@ PREFERRED_VERSION_linux-yocto = "4.16.0%"
 PREFERRED_PROVIDER_virtual/bootloader ?=  "u-boot"
 PREFERRED_VERSION_u-boot ?= "2018.05"
 PREFERRED_VERSION_u-boot-fw-utils ?= "2018.05"
-PREFERRED_VERSION_pulseaudio ?= "10.0%"
-PREFERRED_VERSION_openssl ?= "1.0.2%"
-PREFERRED_VERSION_systemd ?= "234%"
-PREFERRED_VERSION_nodejs = "6.%"
-PREFERRED_VERSION_nodejs-native = "6.%"
+PREFERRED_VERSION_pulseaudio ?= "11.1%"
```
htot: We could remove preferred versions here. Maybe not for node-js.

meta-intel-edison-distro/conf/distro/poky-edison.conf
```diff
@@ -6,17 +6,17 @@ PREFERRED_VERSION_linux-yocto = "4.16.0%"
 PREFERRED_PROVIDER_virtual/bootloader ?=  "u-boot"
```
htot: Changing permissions above is a mistake. I will change to 664 as the other files are.

setup.sh
```diff
@@ -350,7 +350,7 @@ COPYLEFT_LICENSE_INCLUDE = 'GPL* LGPL*'
   # Apply patch on top of it allowing to perform build in external source directory
   echo "Applying patch on poky"
   cd $mingw_dir
-  git apply $top_repo_dir/meta-intel-edison/utils/0001-Enable-SDKTAROPTS.patch
+  #git apply $top_repo_dir/meta-intel-edison/utils/0001-Enable-SDKTAROPTS.patch
```
htot: I found why the patch doesn't apply. Common stuff has been moved to an .inc. So what to do @alext-mkrs ? Is the patch needed? I can fixup the patch so it applies or drop it entirely.

meta-intel-edison-distro/conf/distro/poky-edison.conf
alext-mkrs: Well, I'd actually leave this here and remove e.g. the kernel-ones from the bsp layer's layer.conf (which IMHO is non-intuitive location). Having all preferred versions in one place makes sense to me. Or you mean that after unused layers cleanup these are not necessary? I can double-check that when preparing my PR.

setup.sh
alext-mkrs: TBH I'd drop that Windows SDK thing in its entirety - I'm not sure how well it really works, I don't know how many people use it and finally I believe having a VM for the Linux SDK if necessary is way simpler these days (and avoids all those potential problems).

I don't really know why that specific patch was necessary (it changes layer's defaults) in the first place, so we may as well skip updating it for now.

meta-intel-edison-distro/conf/distro/poky-edison.conf
htot: Eh, I meant do we have a preferred version of f.i. systemd? Or just the latest available?

meta-intel-edison-distro/recipes-core/images/core-image-minimal-initramfs.bbappend
htot: I just tried removing, but then we don't have boot. The reason is that removing this also removes `udev-extraconf` which apparently holds rules for udev.

setup.sh
htot: I already fixed it - but we can drop it any time. That would also require a little more cleanup in the setup script (menu)

htot:
I think I took care of all the comments. Also removed unused mraa, upm and nodejs (from middleware and the meta-nodejs layer).

If you give me the thumbs up I'll start squashing this into a sensible patch set for the 32b noacpi version.

meta-intel-edison-distro/conf/distro/poky-edison.conf
alext-mkrs: Oh, that one I think is yes - just whatever latest we've got, I'm not aware of any specific version dependencies.

htot:
Or maybe it's best to close this PR and reopen a new sumo32 (with reference to this one)?

alext-mkrs:
Nah, I'd vote for just force-pushing, to keep the discussion history in one place.

htot:
Ok. Building now,

alext-mkrs:
I've also run a 32-bit, no-acpi build off of this PR, just to test the configuration closer to `master`. 2123 out of 4274 so far so good.

alext-mkrs:
Yes, builds ok in that configuration.

htot:
OK, with the above clean and squashed commits this now builds no-acpi, 32b version that should be ready to merge. I rebuilt and tested (boots/wifi/bluetooth/i2c) fine. If ok with you I'll merge to master and push a acpi,64b branch.

alext-mkrs:
Yes, lgtm, feel free to merge

htot:
Merged

Issue #28, opened 2018-07-01
On first build kernel modules can not be located for installation into initramfs


htot:
On rocko on the first build after `make setup` the build fails with:
```
ERROR: core-image-minimal-initramfs-1.0-r0 do_rootfs: Unable to install packages. Command '/home/ferry/tmp/edison-intel/my/edison-morty/out/linux64/build/tmp/work/edison-poky-linux/core-image-minimal-initramfs/1.0-r0/recipe-sysroot-native/usr/bin/apt-get  install --force-yes --allow-unauthenticated acpi-tables base-passwd busybox initramfs-live-boot kernel-module-mmc-block kernel-module-sdhci-acpi kernel-module-sdhci-pci run-postinsts udev' returned 100:
Reading package lists...
Building dependency tree...
E: Unable to locate package kernel-module-mmc-block
E: Unable to locate package kernel-module-sdhci-acpi
E: Unable to locate package kernel-module-sdhci-pci

ERROR: core-image-minimal-initramfs-1.0-r0 do_rootfs: Function failed: do_rootfs
ERROR: Logfile of failure stored in: /home/ferry/tmp/edison-intel/my/edison-morty/out/linux64/build/tmp/work/edison-poky-linux/core-image-minimal-initramfs/1.0-r0/temp/log.do_rootfs.10056
ERROR: Task (/home/ferry/tmp/edison-intel/my/edison-morty/out/linux64/poky/meta/recipes-core/images/core-image-minimal-initramfs.bb:do_rootfs) failed with exit code '1'
```
It appears no kernel modules were built. do_compile_kernelmodules was executed, but it's log file says: NOTE: no modules to compile, which seems to originate from grepping .config for "^CONFIG_MODULES=y$" in do_compile_kernelmodules(). Checking .config I find # CONFIG_MODULES is not set which we do have that set if defconfig. Something seems to go wrong with run.do_kernel_metadata on the first run.

run.do_kernel_metadata generates dir's .meta and .kernel-meta under kernel-source. Here the correct config is generated from .defconfig + .cfg. But it seems this is not used by run.do_configure

alext-mkrs:
See also some discussion in #27

alext-mkrs:
[...]

Issue #21, opened 2018-06-01
Is Regulatory DB enabled in the kernel?


lybtongji:
Is Regulatory DB enabled in the kernel?
I want the Edison work in 5GHz AP mode.
Thanks.

htot:
I don't know, I haven't tried. But if the original edison kernel supported it, it should be easy enough to add.

lybtongji:
Thanks for replying.

htot:
We could try to add this. What do you need?

lybtongji:
Thanks. I’m happy to hear that. ^_^

I just want to let the Intel Edison as a Wireless Access Point.

I had let the Intel Edison works in 2.4GHz mode with hostapd, but failed in 5Ghz mode.

Then I found a solution for this problem. link: https://communities.intel.com/thread/57515

It said that the Regulatory DB is not enabled in the original edison kernel.

alext-mkrs:
But as these days we anyway use te upstream kernel, we could enable it right away. Here's the link from that thread  highlighting the option that needs to be enabled: https://communities.intel.com/external-link.jspa?url=http%3A%2F%2Fwireless.kernel.org%2Fen%2Fdevelopers%2FRegulatory%2FCRDA%23CONFIG_CFG80211_INTERNAL_REGDB

htot:
Right, but as I understand the preferred method is to use the crda package. And this will probably just build on rocko: http://layers.openembedded.org/layerindex/recipe/71066/

alext-mkrs:
Yeah, that or the kernel option - in the thread it was mentioned that crda way didn't work right away while kernel option did.

Either way, @lybtongji, if you need that, it could be done.

lybtongji:
Yes, I need it. Thanks very much.

Alex T <notifications@github.com> 于 2018年6月6日周三 00:54写道：

alext-mkrs:
Ok, I'll take a look sometime this week. I'll have to refresh my build setup  so it'll take some time - I'll post back when I have something tangible.

htot:
I think in the thread crda didn't build because edison's yocto was too old. But thanks for taking this @alext-mkrs, on first start of a fresh rootfs I always see annoying errors from hostapd passing by. A reminder that that is still on the todo.

alext-mkrs:
[...]

PR #30, opened 2018-07-08
Rocko kernel build fixes - part 2

alext-mkrs:
Looks like there was more to the issue discussed in #28, so here comes the full fix. There's also further problem with OpenSSL includes (shown in #21), also fixed. While this will hopefully not be necessary after we move to sumo, in the interest of having rocko (and master branch) working, let's have those in.


alext-mkrs:
Thanks, merged. I've tested it extensively, so at the very least it should not break anything :)

Issue #24, opened 2018-06-17
Some scripts use wrong hostapd.conf path


alext-mkrs:
On the current image, the `configure_edison` script uses a wrong path. The file is actually `/etc/hostapd.conf`, but it tries to use `/etc/hostapd/hostapd.conf`. Probably an artifact of an older `hostapd` version that was shipped with original Edison image and needs to be fixed.

ADDED LATER: oh, and `meta-intel-edison/meta-intel-edison-distro/recipes-core/post-install` does that as well.

alext-mkrs:
Okay, this won't be as easy - the recipe is `meta-intel-edison/meta-intel-edison-distro/recipes-support/oobe` and it refers to an external repo: `github.com/01org/edison-oobe`, which is not a major problem to override/patch, but this opens up a topic we've discussed with @htot a while ago, on having our own copies of all dependencies, so that we can preserve/manage them. Looks like this is one more such thing.

@htot, how about creating a copy of this one as well and getting back to our other discussion about moving this out of your personal account to an "organization"? That way we both could have an ability to create repos and I could take care of such things by myself. This should be fairly simple and I can handle it all by myself (create, setup permissions for us both, move repos and update the recipe references) - let's just agree on that being the direction.

htot:
Yeah, I think I already mentioned the postinstall script. Forgot about oobe (actually never realized what the acronym stood for until now).

It's a good idea to create an organisation. So how do I do that?

alext-mkrs:
As organizations are free, I'd propose just a targeted "edison-fw" or "edison-os" to avoid using "intel", to prevent confusion with official repos (and any potential trademark/attribution/endorsement legal problems). Both are unused. And honestly - I believe it almost doesn't matter :) if the contents is good, people will use it anyway, linking is cheap these days and no one remembers full URLs probably anyway.

I'd go with "edison-fw", as IMHO a more generic term, as FW covers both OS and bootloader (and any potential additional stuff we might gather there). Any objections?

htot:
I wanted to create right now edison-fw but is already taken.

alext-mkrs:
Yeah, I went ahead and created it right after the post. I propose a single person (me - describing how to do that will take me longer than doing :) does the initial setup, to avoid misunderstanding and stomping on each other's feet. I'll set us both as owners, but the initial stuff is simply imho more effectively done by one person.

htot:
[...]

PR #31, opened 2018-07-09
Fix hostapd conf path - main

alext-mkrs:
This is the main part of the fix for #24 (the second one is in https://github.com/edison-fw/edison-oobe/pull/1). I've corrected the `hostapd.conf` paths in those scripts and switched to using our copy of the `edison-oobe` repo.

Build-tested on current master.

### DIFF
--- a/meta-intel-edison-distro/recipes-core/post-install/files/post-install.sh
+++ b/meta-intel-edison-distro/recipes-core/post-install/files/post-install.sh
@@ -31,7 +31,7 @@ sshd_init () {
     systemctl start sshdgenkeys
 }
 
-# Substitute the SSID and passphrase in the file /etc/hostapd/hostapd.conf
+# Substitute the SSID and passphrase in the file /etc/hostapd.conf
 # The SSID is built from the hostname and a serial number to have a
 # unique SSID in case of multiple Edison boards having their WLAN AP active.
 setup_ap_ssid_and_passphrase () {
@@ -43,7 +43,7 @@ setup_ap_ssid_and_passphrase () {
         ssid="EDISON-${wlan0_addr:12:2}-${wlan0_addr:15:2}"
 
         # Substitute the SSID
-        sed -i -e 's/^ssid=.*/ssid='${ssid}'/g' /etc/hostapd/hostapd.conf
+        sed -i -e 's/^ssid=.*/ssid='${ssid}'/g' /etc/hostapd.conf
     fi
 
     if [ -f /factory/serial_number ] ;
@@ -52,7 +52,7 @@ setup_ap_ssid_and_passphrase () {
         passphrase="${factory_serial}"
 
         # Substitute the passphrase
-        sed -i -e 's/^wpa_passphrase=.*/wpa_passphrase='${passphrase}'/g' /etc/hostapd/hostapd.conf
+        sed -i -e 's/^wpa_passphrase=.*/wpa_passphrase='${passphrase}'/g' /etc/hostapd.conf
     fi
 
     sync
--- a/meta-intel-edison-distro/recipes-support/oobe/oobe_git.bb
+++ b/meta-intel-edison-distro/recipes-support/oobe/oobe_git.bb
@@ -1,8 +1,10 @@
 DESCRIPTION="The out-of-box configuration service"
 LICENSE = "MIT"
 
-SRC_URI = "git://github.com/01org/edison-oobe.git;protocol=https"
-SRCREV = "4b5f34eed15d15df33af000e474c453ca35245f0"
+SRC_URI = "git://github.com/edison-fw/edison-oobe.git;protocol=https"
+
+SRCREV = "${AUTOREV}"
+PV = "1.2.1+git${SRCPV}"
 
 LIC_FILES_CHKSUM = "file://LICENSE;md5=ea398a763463b76b18da15f013c0c531"
 


htot:
Note: I removed lots of stuff from postinstall, maybe even too much in commit b38c0fe
Also, I think yocto added a postinstall system, and we might consider moving our stuff there and removing this one.
I rebased this PR on top of sumo32. Building resulted in the following warning (harmless?):
`[log_check] warning: TCG doesn't support requested feature: CPUID.80000001H:EDX.lm [bit 29]`

To test I unblocked wifi, disabled wpa_supplicant and started hostapd
```
rfkill unblock 0
systemctl stop wpa_supplicant
systemctl stop hostapd
```
On my phone open network EDISON-4D-70 appears. I believe with the factory image you have a closed network and need to use the serial number as password? When I try to connect it fails after for not receiving a network address.

lybtongji:
You need a DHCP server (like dnsmasq) to give IPs.

alext-mkrs:
So this PR is for fixing just the path itself - as it was breaking the postinstall script (and your earlier cleanup didn't touch that part, which is IMHO good, as it's useful).

Now if the hostapd actually doesn't fully work (at least partially it does - the network is created as you see), that needs to be fixed separately, after this one is merged.

That bitbake warning is a 32bit build artifact I believe (I saw it when testing the sumo swithcing PR), has nothing to do with this change (and it really cannot technically cause it due to the nature of the change). Looks like it's from some qemu log probably being run as a part of building some [of the] package[s]. I believe it's harmless, it apparently says that it doesn't see https://en.wikipedia.org/wiki/Long_mode capability in CPUID and as we're building 32bit image, that's expected.

alext-mkrs:
Oh and the lack of passphrase is indeed because of that cleanup, now I see it :)

`/factory/serial_number` stored the serial number and it got removed in https://github.com/edison-fw/meta-intel-edison/commit/b38c0fec27797b906593091fcb26badb6c526758#diff-6ae83b99829fe64a9082facc28b8538bL70. But we can correct that later.

htot:
@alext-mkrs Thanks, I didn't mean to criticize, just to summarize as a note to self what I tested and what could be added later.

The patch does what it advertises. So let's go ahead and merge.

@lybtongji Thanks too, unfortunately we currently don't have dnsmasq on the image. I built the deb and installed manually and tried your command. Unfortunately it didn't work. I noticed it creating a systemd dnsmasq.service. And we also have systemd-networkd. Now trying to figure out how that works.

I would like to know what you guys would like to use hostapd for. For me:
1) I have a usb->ethernet port connected. It would be cool to bridge that to wifi.
2) In a iot application, with redundant edisons it would be nice to have peer-to-peer or one hostap / one client as a fallback when ethernet fails.

alext-mkrs:
@htot, thanks for the review. No offense taken! :) On the contrary, I really appreciate that you flag things you notice - that's the way for us to find them and fix.

As for the `hostapd` use cases - to be honest I personally have none, so I'm open to all the suggestions/asks people may have. One that was used in the offical image was that initial configuration - out of the box Edison starts in AP mode, so a user can connect to it and configure basic stuff using web interface (as opposed to manually running `configure_edison` in the console, which is what I used to do, or just doing plain config file editing). We can collect those use cases and try to implement.

I'll merge this one and then look into why the address is not given out, I'd expect it to. Probably something else got swcrewed up along the way or changed behavior due to package version changes. :) The basic idea I have right now is to get back the official image functionality I described above. After that we could look into other use cases and see what it takes to implement them.

htot:
@lybtongji BTW I managed a DHCP server in the AP. First removed dnsmasq., the created `20-wlan-dhcp.network` in `/etc/systemd/network` with:
```
[Match]
Name=wlan*

[Network]
#Bridge=br0
DHCPServer=yes
PoolOffset=192.168.1.10
PoolSize=20
Address=192.168.1.1/24
Gateway=192.168.1.1
```
With this I can connect my phone (as long as it doesn't check the 'network quality' as the gateway doesn't work).
I can't create a bridge (yet):
`br0: netdev could not be created: Operation not supported`
I think we need to enable kernel support for this first?
And @alext-mkrs I was thinking the same on the config server. I think it worked with a web server in nodejs. But where is that? Maybe I removed that in the past, with out much consideration.

lybtongji:
@htot I have no problem with `dnsmasq` in `rocko32`.
After edison boot into system, I unblock the `wlan0` by
``` sh
rfkill unblock all
```
Then I run `dnsmasq` in `screen`:
``` sh
screen -S dnsmasq
dnsmasq -k -p 0 -x /var/run/dnsmasq.pid -i wlan0 -F wlan0,192.168.1.2,192.168.1.10,12h -O wlan0,3,192.168.1.1
```
Then start `hostapd` in another `screen`:
``` sh
screen -S hostapd
hostapd hostapd.conf
```
Finally, I will see an AP named "test" without password. And I can connect to it directly. IP will be assigned automatically.

htot:
I got `node /usr/lib/edison_config_tools/edison-config-server.js` to work. There are multiple problems:
1) iwlist is installed, but probably compatibility in the kernel is not enabled
2) the modern replacement is not installed in the image
3) manually installing, the output is different, requiring changes to configure_edison
4) shelljs used to have output field, but changed to stdout

I will open a new issue for this.

alext-mkrs:
Merged.

lybtongji:
@htot After I added `CONFIG_BRIDGE=m` to kernel conf file and rebuilt the `rocko32` image, the bridge mode in `hostapd` is working fine.

``` sh
root@edison:~# brctl show
bridge name     bridge id               STP enabled     interfaces
br0             8000.90b6860de3b2       yes             wlan0
```

PR #43, opened 2018-08-15
Sumo64 acpi

htot:
This should include the sumo32 changes + when changing to acpi also enables acpi on kernel command line.


htot:
Thanks. That was a 'x' instead of <ctrl-x>.
The revert 32b, reverts back to 64b, not very clear but actually correct for this branch.

htot:
Fixed both signed off lines

alext-mkrs:
Looks ok, to me now. Indeed, I skimmed through the 32bit commit too fast and took it for the opposite one :) sorry.

htot:
No worries.

PR #47, opened 2018-09-30
Feature reboot arg

### LINKED ISSUE #45: Since morty `reboot arg` does not work and just reboots
This is still an issue on sumo.
Factory Edison image bbappends systemd to achieve this.

The use case would be to reboot into edsboot from ssh using `reboot sd` and usbboot using `reboot usb`.

While here, we might remove `reboot ota` as that can't work due to the size of the image and the limited free space. Maybe we can  find an alternative when extra storage (sd/usb) is connected, or update via `apt`.

htot:
This commit allows you to type `reboot usb` and `reboot sdhc`, which will perform `usbboot` or `edsboot` respectively without the need to interrupt U-Boot from the console.

The advantage is you can reboot from ssh from remote into the image you want (without console).

Other supported bootargs are `boot`, `ota` and `flash` (unchanged from the factory image).

This fixes #45

### DIFF
--- a/meta-intel-edison-bsp/conf/machine/edison.conf
+++ b/meta-intel-edison-bsp/conf/machine/edison.conf
@@ -6,7 +6,7 @@
 # This sets compilation options close to what is used on android
 include conf/machine/include/tune-core2.inc
 TUNE_CCARGS .= " -mstackrealign"
-#DEFAULTTUNE = "core2-64"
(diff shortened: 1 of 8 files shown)


alext-mkrs:
Just so you know - I have this on my radar for the upcoming weekend

htot:
Merged

Issue #44, opened 2018-08-20
On sumo64-acpi blink-led is not working


htot:
@lybtongji wrote:  
> I flashed sumo64-acpi and found blink-led not working. Is it due to ACPI?

@htot wrote:
>Yes, because of acpi pin names/numbers have changed. This affects blink-led and mraa and probably some other code too. I found the best way to access pins is from the libgpiod provided tools gpiodetect gpiofind gpioget gpioinfo gpiomon gpioset. See also Edison wiki.
For now I think we need the code first to detect if we are using acpi (Edit: /sys/kernel/config/acpi/table/arduino/ /sys/firmware/acpi exists).
I didn't find time to fix blink-led, but if you want to take this, please do.

htot:
@lybtongji I created a new issue for this.

lybtongji:
Ah. It just a question, not very necessary. But I'm glad to hear that you can help with it.

In addition. Can I make the LED (`DS2`) to be an activity light of `CPU` or `eMMC`?

which I found in kernel configurations:

```
...
CONFIG_LEDS_TRIGGER_DISK
CONFIG_LEDS_TRIGGER_ACTIVITY
...
```

htot:
The blink-led program is probably relatively easy to fix. Also you should be able to turn LED on and off from command line, You need to figure out the name of the pin, try `gpioinfo`.
To turn on: gpioset `gpiofind pin_name`=1

I haven't studied `CONFIG_LEDS_TRIGGER_DISK` and `CONFIG_LEDS_TRIGGER_ACTIVITY`, so I don't have the answer to that.

htot:
Fixed with PR #48

PR #48, opened 2018-10-13
Fix blink led

htot:
This fixes issue #44 (on the acpi enabled kernel pin numbering has changed breaking the blink-led program).

On acpi the preferred way to access gpio is through libgpiod. The blink-led script is python2.7. libgpiod supports python3 bindings, however the sumo provided recipe does not enabled this.
1) get libgpiod from master
2) enabled python3 bindings
3) spi hogs a pin needed for blinking the led, disable spi for now
4) fixup blink-led to run with python3 and switch to using libgpiod when acpi is detected
5) remove python packages from initramfs as we exceed maximum size

I tested this both with and without acpi enabled kernel.

In a later stage I intend to enable kernel led driver for acpi enabled image, with led trigger (timer, heartbeat, morse) which will probably obsolete blink-led as a service. Until then this works and documents how to migrate from sysfs to libgpiod (which we need to do for mraa as well - which is the major obstacle to make acpi enabled image the default).

### DIFF
--- a/meta-intel-edison-distro/recipes-core/images/core-image-minimal-initramfs.bbappend
+++ b/meta-intel-edison-distro/recipes-core/images/core-image-minimal-initramfs.bbappend
@@ -12,9 +12,13 @@ PACKAGE_INSTALL += " ${@bb.utils.contains('DISTRO_FEATURES', 'acpi', 'acpi-table
 PACKAGE_INSTALL_remove = " initramfs-live-install initramfs-module-install"
 PACKAGE_INSTALL_remove = " initramfs-live-install-efi initramfs-module-install-efi"
 PACKAGE_INSTALL_remove = " kernel-image"
+PACKAGE_INSTALL_remove = " libgpiod-python python3-core libpython3.5m1.0"
 
 ROOTFS_POSTPROCESS_COMMAND += "clobber_unused; "
 
 clobber_unused () {
         rm ${IMAGE_ROOTFS}/boot/*
+# libgpiod2 pulls in python3-core which we don't need
+        rm -r ${IMAGE_ROOTFS}/usr/lib/python3.5/*
+        rm ${IMAGE_ROOTFS}/usr/lib/libpython3.5m.so.1.0
 }
--- a/meta-intel-edison-distro/recipes-support/blink-led/files/blink-led
+++ b/meta-intel-edison-distro/recipes-support/blink-led/files/blink-led
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 #
 # Edison LED blinker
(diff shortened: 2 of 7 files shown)


lybtongji:
:smile:

meta-intel-edison-distro/recipes-support/blink-led/files/blink-led
```diff
@@ -73,29 +91,43 @@ args = parser.parse_args()
...
+    echo 243 >/sys/class/gpio/export
+    echo 261 >/sys/class/gpio/export
+    echo high >/sys/class/gpio/gpio214/direction
+    echo low >/sys/class/gpio/gpio243/direction
+    echo high >/sys/class/gpio/gpio261/direction
+    echo low >/sys/class/gpio/gpio214/direction""", shell=True)
+
+    # Save current LED value for reverting to proper state at exit
+    try:
+        lf =  open("/sys/class/gpio/gpio40/value","r")
```
alext-mkrs: Surplus space after "="

meta-intel-edison-distro/recipes-core/images/core-image-minimal-initramfs.bbappend
```diff
@@ -12,9 +12,13 @@ PACKAGE_INSTALL += " ${@bb.utils.contains('DISTRO_FEATURES', 'acpi', 'acpi-table
 PACKAGE_INSTALL_remove = " initramfs-live-install initramfs-module-install"
 PACKAGE_INSTALL_remove = " initramfs-live-install-efi initramfs-module-install-efi"
 PACKAGE_INSTALL_remove = " kernel-image"
+PACKAGE_INSTALL_remove = " libgpiod-python python3-core libpython3.5m1.0"
 
 ROOTFS_POSTPROCESS_COMMAND += "clobber_unused; "
 
 clobber_unused () {
         rm ${IMAGE_ROOTFS}/boot/*
+# libgpiod2 pulls in python3-core which we don't need
```
alext-mkrs: This looks like wrong indentation

meta-intel-edison-distro/recipes-support/blink-led/files/blink-led
```diff
@@ -49,18 +59,26 @@ def deinit_led_gpio():
...
+        subprocess.call("""
+        echo 40 >/sys/class/gpio/unexport
+        echo 214 >/sys/class/gpio/unexport
+        echo 243 >/sys/class/gpio/unexport
+        echo 261 >/sys/class/gpio/unexport
+        """, shell=True)
 
 def signal_term_handler(signal, frame):
-    print 'Signal intercepted: de-initing LED GPIOs'
+    print('Signal intercepted: de-initing LED GPIOs')
```
alext-mkrs: Throughout the script single and double-quoted strings are intermixed. I'd suggest to stick to one style or the other for uniformity.

alext-mkrs:
> and documents how to migrate from sysfs to libgpiod (which we need to do for mraa as well - which is the major obstacle to make acpi enabled image the default)

Do you mean that mraa doesn't expect to see newer kernels on Edison, so doesn't use the new interface? This can be fixed, yes. The only thing is that mraa doesn't use libgpiod, it uses the new chardev interface directly, but OTMH this shouldn't be a problem.

htot:
@alext-mkrs thanks for reviewing. I'll resolve your comments and merge,

On mraa: you explained before that mraa doesn't want dependency on libgpiod but I don't think the chardev interfaces is used yet on eds. We are on mraa 1.9 and afaiu [intel_edison_fab_c.c](https://github.com/intel-iot-devkit/mraa/blob/master/src/x86/intel_edison_fab_c.c) is used which uses sysfs.

htot:
merged

alext-mkrs:
Understood. Yes, mraa can be updated similar to other platforms supported, which have this functionality.

PR #52, opened 2018-11-01
initramfs: check python packages prior to remove

### LINKED ISSUE #51: When acpi is not set clobber fails because python not found
https://github.com/edison-fw/meta-intel-edison/blob/a41f2a7d93e1e595ab682447639e94406cfe97a0/meta-intel-edison-distro/recipes-core/images/core-image-minimal-initramfs.bbappend#L22

Look like this needs to be either conditional on acpi or on the presence of the directory

htot:
This fixes #51

### DIFF
--- a/meta-intel-edison-distro/recipes-core/images/core-image-minimal-initramfs.bbappend
+++ b/meta-intel-edison-distro/recipes-core/images/core-image-minimal-initramfs.bbappend
@@ -19,6 +19,7 @@ ROOTFS_POSTPROCESS_COMMAND += "clobber_unused; "
 clobber_unused () {
         rm ${IMAGE_ROOTFS}/boot/*
         # libgpiod2 pulls in python3-core which we don't need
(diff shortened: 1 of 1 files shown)


alext-mkrs:
So I think we don't really need that logic in place - `rm -rf` should do the trick. Sorry for accidentally misleading you in #51 - I haven't noticed the `-f` didn't actually make it into my comment.

htot:
Yeah, no problem. I think I already created this patch before I read your comment - and then was to lazy to change it. Mostly because I was wondering if -f removes the error condition, I thought it meant 'force'. But I 
see now and will change this, simple enough.

htot:
This build without errors and the size of the compressed cpio is 3.2MB same as with my V1. Thanks for the suggestion.

htot:
Merge it?

alext-mkrs:
Yep

htot:
merged, thanks

PR #53, opened 2018-11-01
arch: switch to corei7

htot:
This fixes an old question you raised:
Edison is known to support corei7 instructions. On Pyro setting tune to corei7-32 caused non-
functioning sdk build.

But on Sumo sdk works fine so switch corei7. Being a slm atom core it is better optimized for 32bits.  Switch tune for sumo32.


htot:
Thanks, merged.

Issue #50, opened 2018-10-14
MRAA needs to be adapted to acpi pin numbering/naming


htot:
When we run the MRAA Example Blink-IO.js on the acpi enabled image nothing happens.

dmesg shows:
`edison kernel: export_store: invalid GPIO 214`

This indicates MRAA is using the wrong pin numbering - and in fact it should be using the pin names.
Doing `gpioinfo` shows `TRI_STATE_ALL` is on `gpiochip1` line 14.
You find `gpiochip1` under `/sys/class/gpio/gpiochip496/device/`, so the base address has become 496 and the gpio is 510.

We need to modify [intel_edison_fab_c.c](https://github.com/intel-iot-devkit/mraa/blob/master/src/x86/intel_edison_fab_c.c) to use either libgpiod or the new gpio chardev interface directly to get mraa working. 

This is a priority task as it currently seems to be the only thing holding us back to switch to the acpi enabled image.

@alext-mkrs do you feel inspired to take this? You are without doubt the most knowledgeable of us.

alext-mkrs:
I'm sorry - I've apparently [dis]missed the notification for this one. I think I could take that on and being the mraa co-maintainer is sort of fitting here :D The only problem is that I haven't done much actual development there for a while, so will need to refresh myself on platform internals + find the time to actually do that - and the latter is the main problem these days.

Anyway, let me assign this to myself and see what (and when) I can do - I'll post the progress here.

To start at the beginning - what's the latest and greatest ACPI image build recipe, so that I have a live platform to work off of?

htot:
@alext-mkrs take would be just great. Latest and greatest is here https://github.com/htot/meta-intel-edison .. since yesterday. But I didn't dare to ask you yet.

I sort of EOL the sumo branches, and of course immediately sumo itself got updated and broke the build. With yesterdays patches it builds again on 18.10 Cosmic. I'm hoping someone (@sjpeng) will build those to verify they still build on earlier (like 18.04). And then PR to edison-fw. 

But for this effort (MRAA) the brand new thud branch (64b) or master (32b) would be nice, 4.20 kernel would be best. 
Tested yesterday, all seems to be working well. USB OTG (gadget) works too, but I only tested with linux hosts. I guess for windows more needs to be done, windows needing drivers and inf files. Docs also in master and therefore here: https://htot.github.io/meta-intel-edison/ (see the USB section).
[...]

PR #57, opened 2019-01-19
Revert me: patches to make build on Ubuntu Cosmic 18.10

### LINKED ISSUE #56: sumo branch build failure
Build Configuration:
BB_VERSION           = "1.38.0"
BUILD_SYS            = "x86_64-linux"
NATIVELSBSTRING      = "universal"
TARGET_SYS           = "i686-poky-linux"
MACHINE              = "edison"
DISTRO               = "poky-edison"
DISTRO_VERSION       = "2.5.2"
TUNE_FEATURES        = "m32 corei7"
TARGET_FPU           = ""
meta
meta-poky
meta-yocto-bsp       = "sumo:623b77885051174d0e05198843e739110977bd18"
meta-oe
meta-python
meta-networking      = "sumo:8760facba1bceb299b3613b8955621ddaa3d4c3f"
meta-intel           = "sumo:90af97d23fb2a56187c2fe2a3f4f4190d7cc2605"
meta-intel-edison-bsp
meta-intel-edison-distro = "master:d0de39d612c6cb032f19acd3b89f2f4264c444b9"
meta-acpi            = "eds:477a31d2bad06de9c240753462695c46569cf162"

ERROR: Multiple versions of openssl are due to be built (/home/cloud/edison/out/linux64/poky/meta/recipes-connectivity/openssl/openssl_1.1.0i.bb /home/cloud/edison/out/linux64/poky/meta/recipes-connectivity/openssl/openssl_1.0.2p.bb). Only one version of a given PN should be built in any given build. You likely need to set PREFERRED_VERSION_openssl to select the correct version or don't depend on multiple versions.

Summary: There was 1 ERROR message shown, returning a non-zero exit code.
Makefile:53: recipe for target 'edison-image' failed
make: *** [edison-image] Error 1

htot:
These patches are needed to build on Ubuntu Cosmic due to glibc 2.28. I tested this on Cosmic and on Xenial to verify backwards compatibility. It has been independantly tested by @sjpeng. I expect at some point these patches will be backported to Sumo at which point this will break and need to be reverted.

This closes #56


meta-intel-edison-distro/conf/distro/poky-edison.conf
```diff
@@ -7,7 +7,6 @@ PREFERRED_PROVIDER_virtual/bootloader ?=  "u-boot"
 PREFERRED_VERSION_u-boot ?= "2018.09"
 PREFERRED_VERSION_u-boot-fw-utils ?= "2018.09"
 PREFERRED_VERSION_pulseaudio ?= "11.1%"
-PREFERRED_VERSION_openssl ?= "1.1.0%"
```
alext-mkrs: I think this one needs a closer and a separate look - as this should have nothing to do with the glibc problems other patches in the PR solve. And downgrading the OpenSSL doesn't look good to me. Let me try to look into this while I'm prepraring a build for #50.

htot: No, apparently something depends on openssl 1.0.2p. Which it didn't before.

alext-mkrs: Yeah, I understood from your discussion in #56 that's something unexpected and possibly related to sumo updates, however it's a bit hard to believe they'd change something in favor of downgrading the OpenSSL, so I think that may be some bug/misconfig somewhere we could chase.

htot: So what do think? 

BTW if we can pull this then we have a more or less stable sumo. The new thud branch that I will send a PR for next has openssl 1.1 really. And U-Boot 2019.01, kernel 4.20. Is much more interresting :-)

alext-mkrs: Just got to investigate that - right, what you've outline is exactly the reason. Let's then just go with removing the preferred version indeed for now and see if `thud`'s packages no longer require 1.0. 1.1.1 is now the LTS version and has good security-related updates across the board, so I'd be interested in updating to that, but that's not really a critical piece for our image anyway :)

htot: Yes thud has 1.1.1a see https://github.com/htot/meta-intel-edison/blob/thud/docs/edison-image-edison.manifest

htot:
Thanks for reviewing.

PR #58, opened 2019-01-27
Move on to Thud

htot:
This is a large upgrade with notable changes:
- Acpi: remove switch to build non-acpi image
- Poky: switch to Thud
- U-Boot: 2019.01
- Linux: 4.20
- Gadget: enable support for eth over usb
- Timezone: detect and set timezone
- Ap-mode-toggle: fix it
- Bluez: build options to allow pan (unfortunately I didn't get pan working yet)
- Python3-bottle: prepare for future fix of oobe
- Docs: document above changes + new pages on container / repo / Flashtool Lite

Not yet included:
- enable LED subsystem to blink the led
- oobe re-write (will be based on bottle and remove the nodejs dependency)


alext-mkrs:
Will check it out, some nasty flu knocked me out for the whole last week, just getting back on track.

htot:
Get well!

kobylyanskiy:
Hi @htot ,
Thank you for your work on Edison. I'm currently checking out your commits and have built new image. But since I'm not using connman, I want to know how to enable networking over USB on the lower level rather than using ```connmanctl```. ```nmcli dev status``` says that usb0 connection is up, but I'm getting no luck to ping the device. It would be great if you could say whether it's a problem with powering on the gadget or something else.

htot:
@kobylyanskiy If I understand correctly you installed networkmanager instead of connman and already read https://htot.github.io/meta-intel-edison/4.2-networking.html#gadget? (Note: I see now this section is a little broken, net1, net2, and net3 are placeholders for images. I'll fix that later)

My best guess is indeed powering on the gadget. Connman can also power on bluetooth and wifi as an alternative to rfkill. I'm not sure to how to power the gadget without connman. Maybe nmcli has similar functionality?

htot:
@kobylyanskiy maybe this will power up: `ip link set usb0 up`

If that doesn't work, how are you configuring your host side? Do you see a usb device popping up as well?

htot:
Heads up: I force pushed to fix the docs on gadget networking.

alext-mkrs:
Looks generally okay to me by quick inspection. I'll test this out by building right away, as I anyway need a build for the mraa porting.

htot:
There's some rough edges here and there. F.i. the udev script that creates the gadget should be powering up usb0, but it's too early. It needs a delay, or a 2nd udev event. Working on that.

htot:
@alext-mkrs thanks for reviewing. I squashed that last commit and force pushed.

PR #59, opened 2019-04-14
LInux 5.0.0, U-Boot 2019.04

htot:
This PR brings in latest released kernel + U-Boot and
- latest ACPI tables from meta-acpi (eds-5.0.0) (see [README](https://github.com/edison-fw/meta-acpi/tree/eds-5.0.0) for the list of supported tables)
- enables kernel heart beat led in favor of user space blinkled service
- adds hardware rng to provide sufficient entropy needed to login in fast over ssh
- enables bpf to shutup systemd complaining
- set bluetooth hw address from /factory, allows bt tethering from your android phone or laptop
- powers gadget ethernet by default (provides 200Mb/s ethernet over usb to a linux host, haven't tried with windows yet)
- updated [docs](https://htot.github.io/meta-intel-edison/)


htot:
@alext-mkrs would you have time to review this?

htot:
@alext-mkrs do you think you will be able to find time to look at this?

alext-mkrs:
Yeah, I think I will - in the nearest couple of days.
Sorry, it's been tough to find the time for the last several months for anything... but I'm still trying to keep up with the repo changes :)

htot:
No problem, I wanted to finish oobe before Feb, but it's still not done :-)

meta-intel-edison-distro/recipes-core/post-install/files/post-install.sh
```diff
@@ -31,20 +31,40 @@ sshd_init () {
...
+setup_ap_ssid_and_passphrase () {
+    # factory_serial is 16 bytes long
+    if [ -f /sys/class/net/wlan0/address ];
+    then
+        wlan0_addr=$(cat /sys/class/net/wlan0/address | tr '[:lower:]' '[:upper:]')
+        ssid="EDISON-${wlan0_addr:12:2}-${wlan0_addr:15:2}"
+    fi
 
-# script main part
+    if [ -f /factory/serial_number ] ;
```
alext-mkrs: This one IIRC was populated by one of the first-boot scripts, so theoretically may not exist everywhere or may be deleted by the user. If there's no such script in place right now (to guarantee the serial will always exist), it's probably worth it to use some fallback values for the SSID and password to prevent  creating invalid or passwordless config.

meta-intel-edison-bsp/recipes-kernel/bcm43340/bcm43340-addr_0.1.bb
```diff
@@ -0,0 +1,29 @@
+DESCRIPTION = "Daemon listening to Edison PWR long button press, and starting OOBE service when it happens"
```
alext-mkrs: That's probably not the description you wanted to have for this one. Copy artifact?

meta-intel-edison-distro/recipes-core/post-install/files/post-install.sh
htot: Sure? The directory contains 2 files with serial number equal to the label on the edison and the other a unique bluetooth address. I was under the impression the files are created at production time and never overwritten during flashing (or I would have lost them long ago). Nevertheless it is good advice to have some defaults just in case. I'll have a look.

meta-intel-edison-bsp/recipes-kernel/bcm43340/bcm43340-addr_0.1.bb
htot: Yes, thanks for finding.

meta-intel-edison-distro/recipes-core/post-install/files/post-install.sh
alext-mkrs: I think [this function](https://github.com/edison-fw/meta-intel-edison/commit/b38c0fec27797b906593091fcb26badb6c526758#diff-6ae83b99829fe64a9082facc28b8538bL70) is the source for my impression, but I can't find any code that actually populates those files with meaningful data, so probably that dummy data in the function is a fallback similar to what I've proposed in this comment.

htot: Ah, I forget about that one. Yes, I didn't find flashall writing that factory partion either. I guess in the factory they had a netboot over rndis or cdc, generate serial and pick bd_addr from database, then put label with serial. But once you loose the partition, there is no way to recover it.

htot: I had a closer look at the original code, and as I understand now on first run /factory is created and mounted, only if the mount fails a partition is created and defaults are entered. As the code says, that can happen in the factory. But why don't we now error out? We are not creating /factory anymore. Is this taken care of by systemd due to our line in fstab:
```
/dev/disk/by-partlabel/factory  /factory        auto    noauto,x-systemd.automount,nosuid,nodev,noatime,discard     1   1
```

alext-mkrs: Maybe that's because none of the systems anyone used have actually lost their partitions and the reflashing process being used doesn't touch the existing ones? Alternatively - people haven't noticed :)

htot: I meant to say: we produce a new root image, afaik without /factory directory. Still we are able to mount the factory partition onto it. So, who creates the /factory directory?

htot:
@alext-mkrs so I fixed up the code according to your remarks and added 2 docs commits. If you agree I will pull. Thanks!

PR #62, opened 2019-06-29
Improve build time for CPU with more than 4 cores.

ochrin:
Small fix which let Yocto decides how many threads are needed for building Yocto. It will be normally be set to number of cores.
If `SETUP_ARGS` is not defined with either one or both of these parameters `--bb_number_thread=x --parallel_make=y` when calling make setup, Yocto will set itself the number of threads to be used and how many cpu to be used for parallel makes.
In my case with 16 cores, with the old setup.sh (4 threads, 8 cpus for parallel make) I needed 106mn to build the image with bitbake, with new setup.sh, I need 73mn.


htot:
Since htot was ahead of edison-fw 5 additional commits were also pushed here. I want to defer those until I have oobe ready and all can go in after review.

56bfa51 LGTM by inspection. Nice improvement in speed to. Probably there will even be a speed improvement on dual cores due to better matching of threads with cores (when using the defaults).

PR #66, opened 2019-08-27
Update 4.4-hsu-i2c-spi.md

ecc1:
Describe results of testing SPI via the /dev/spidev5.1 device.


htot:
Thanks. I guess a next step would be to enable a kernel driver. Not sure if that already made it into the kernel. I found this patch https://lore.kernel.org/lkml/826180465.66159.1500124248340@ox.hosteurope.de/#t

htot:
Merged, thanks.

PR #67, opened 2019-09-15
Preparing for EOL thud

htot:
I added stuff to pair bluetooth and upgraded kernel. I didn't get to finishing oobe (which is in it's independent repo), so the web service to control all these features is not yet in place. Nevertheless, everything is prepared for that. We can enable (over the command line using a terminal and connmanctl) wifi connection, wifi AP, bluetooth connection, eth over usb.

So, I would like to get this in and EOL thud. Then switch to warrior.


meta-intel-edison-bsp/recipes-support/pwr-button-handler/files/simple-agent
```diff
@@ -0,0 +1,213 @@
...
+	err_name = error.get_dbus_name()
+	if err_name == "org.freedesktop.DBus.Error.NoReply" and device_obj:
+		print("Timed out. Cancelling pairing")
+		device_obj.CancelPairing()
+	else:
+		print("Creating device failed: %s" % (error))
+	mainloop.quit()
+
+def get_pin():
+	f = open("/factory/serial_number", "r")
```
alext-mkrs: It's generally considered a bad practice to leave opened files to a garbage collector. I'd suggest to use either an explicit `close()` or `with`. I'm not sure though if you're just copying this file from somewhere (e.g. old Edison distro or something)?

PR #70, opened 2019-09-30
added support for OTA from pocky to thud and from thud to thud

razvan-becheriu:
added support for OTA from pocky to thud and from thud to thud by updating boot environment


razvan-becheriu:
this MR must be done after merging https://github.com/edison-fw/u-boot/pull/1

razvan-becheriu:
Downgrading form thud to pocky is not supported, as it needs a similar patch applied to the pocky version.

PR #71, opened 2019-09-30
added support for OTA from pocky to thud and from thud to thud

razvan-becheriu:
added support for OTA from pocky to thud and from thud to thud by updating boot environment


razvan-becheriu:
this MR must be done after merging edison-fw/u-boot#1

razvan-becheriu:
Downgrading from thud to pocky is not supported, as it needs a similar patch applied to the pocky version.

htot:
This went in as e420040f229e5ddd8abc9fad24fc111e88944219 thanks.

See also https://github.com/edison-fw/u-boot/pull/2

PR #79, opened 2019-11-29
Update 2.3-Building-and-flashing-U-boot.md

mwallnoefer:
fix dead link + remove spaces between partition entries

### DIFF
--- a/docs/_docs/Edison/Building/2.3-Building-and-flashing-U-boot.md
+++ b/docs/_docs/Edison/Building/2.3-Building-and-flashing-U-boot.md
@@ -22,7 +22,7 @@ See:
 
 The status is being tracked here: 
 
-* [https://edison.internet-share.com/w/index.php?title=Using_a_vanilla_Linux_kernel_with_Intel_Edison&redirect=no](https://edison.internet-share.com/w/index.php?title=Using_a_vanilla_Linux_kernel_with_Intel_Edison&redirect=no).
+* [https://edison-fw.github.io/edison-wiki/vanilla](https://edison-fw.github.io/edison-wiki/vanilla).
 
 ### Instructions for building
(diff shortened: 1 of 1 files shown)


docs/_docs/Edison/Building/2.3-Building-and-flashing-U-boot.md
```diff
@@ -137,11 +137,11 @@ You can change a line by using `setenv/fw_setenv`.
 *   The following lines need to be changed since the keyword mmc has changed to raw. For example inside u-boot:
```
andy-shev: If the original configuration has spaces it would make sense to update text above to reflect what you described in the commit message.

mwallnoefer: No, I don't think that it had spaces - this was a mistake in this particular file.

I have checked also other sources (Intel Forum etc. https://forums.intel.com/s/question/0D50P0000490T69SAE/update-uboot?language=en_US) and nowhere there were spaces between. It took me also quiet some time to realise that this was the issue.

htot: @mwallnoefer you are right, the error is in de documentation only, and applies to the instructions for manually fixing the environment. When performing flashall, the environment is also reflashed to the one generated by [this file](https://github.com/edison-fw/meta-intel-edison/blob/master/meta-intel-edison-bsp/recipes-bsp/u-boot/files/edison.env), which is correct without spaces.

htot:
I rephrased a bit and push to https://github.com/htot/meta-intel-edison (master and warrior). Back ported onto https://github.com/edison-fw/meta-intel-edison (thud32-v2.0 and thud).

Eventually this will land into edison-fw/meta-intel-edison master.

Thanks.

PR #89, opened 2020-02-22
edison-machine-id.service - drop the umount

mwallnoefer:
It always generates an error like
> umount: /etc/machine-id: not mounted.


htot:
If all is as I expect systemd already generates an id, so we may drop this recipe entirely. Let me test that and get back.

PR #90, opened 2020-02-22
edison-image.bb - disable battery-voltage.service

mwallnoefer:
May only be re-enabled if the pmic_ccsm service gets restored (consider battery-voltage/battery-voltage/battery-voltage.c).

### DIFF
--- a/meta-intel-edison-distro/recipes-core/images/edison-image.bb
+++ b/meta-intel-edison-distro/recipes-core/images/edison-image.bb
@@ -92,8 +92,8 @@ IMAGE_INSTALL_append = " mcu-fw-bin"
 # Add oFono
 IMAGE_INSTALL_append = " ofono"
 
-# Add battery level detection
-IMAGE_INSTALL_append = " battery-voltage"
+# Add battery level detection (disabled since 'pmic_ccsm' not available anymore)
+#IMAGE_INSTALL_append = " battery-voltage"
 
 # Add linux performance monitoring tool
 IMAGE_INSTALL_append = " perf"


andy-shev:
pmic_ccsm driver will never be upstreamed, So, if battery service at some point will be revived, it will be (completely?) different approach. So, I would go further and remove this from distribution for time being.

htot:
afaikt all this service does is get the battery voltage and print it. So now, from bcove_adc we get
`cat /sys/bus/platform/drivers/mrfld_bcove_adc/mrfld_bcove_adc/iio:device0/in_voltage0_raw`
returns 988, which probably corresponds to 988 / 1024 * 4.5 = 4.3V.

So, what does pmic_ccsm do extra?

andy-shev:
> afaikt all this service does is get the battery voltage and print it. So now, from bcove_adc we get
> `cat /sys/bus/platform/drivers/mrfld_bcove_adc/mrfld_bcove_adc/iio:device0/in_voltage0_raw`
> returns 988, which probably corresponds to 988 / 1024 * 4.5 = 4.34V.

Yes, monitoring (PMIC ADC) works...

> So, what does pmic_ccsm do extra?

...charging is not. 

Sorry for the confusion, because I didn't quite get what is that service for.

htot:
> Sorry for the confusion, because I didn't quite get what is that service for.

Neither do I. It seems to print battery status to syslog? But why?

htot:
Anyway I fixed it, useless as it seems. Will appear initially in https://github.com/htot/meta-intel-edison/tree/zeus

xlla:
I saw this in every boot log.
<pre>
[FAILED] Failed to start Battery Voltage daemon.
See 'systemctl status battery-voltage.service' for details.
</pre>

andy-shev:
So, charging actually works on Sparkfun Battery block and on Edison/Arduino according to the data sheets of certain discrete components. The idea behind the service is probably to shutdown correctly the machine, or do something else.

andy-shev:
Okay, I investigated battery level mechanism. It seems that the above service just prints it in human readable format, while the formula is indeed `(x) * 450 / 1024 = Battery Level in centivolts` (`pmic_get_battery_voltage()` in the stock kernel sources), where `(x)` is the 16-bit value from Channel 0, i.e. `VBATRSLT`. So, this service is doubtfully needed.

Thus, I think removal is the best option to avoid user confusion.

htot:
I created a branch zeus and created a patch to fix battery-voltage.service as suggested here. Thanks.

PR #92, opened 2020-02-22
fstab - mount /boot by using the partition label

mwallnoefer:


htot:
I created a branch [zeus](https://github.com/htot/meta-intel-edison/tree/zeus) and pulled this there. Thanks.

PR #99, opened 2020-03-14
4.2-networking.md - passphrase

mwallnoefer:


mwallnoefer:
Please don't forget this one

htot:
Actually, [post-install ](https://github.com/edison-fw/meta-intel-edison/blob/6bc6d5f9de0599f5c26f86406f19b423a6fe0a69/meta-intel-edison-distro/recipes-core/post-install/files/post-install.sh#L37) sets the passphrase.

So why the minimum of 8?

mwallnoefer:
It needs to be at least 8 chars long for the WPA2 standard?

htot:
Is that a requirement from the standard (I didn't know) or from certain access points. I see that hostapd has a minimum of 8. Can't find a reference in the connman sources.

mwallnoefer:
I think that it is required by the standard, also `configure_edison` had it hard-coded. And I have already tried with a shorter password where I ended up in an error.

htot:
Alright. Did I read correct you dropped support wps and open networks?

mwallnoefer:
Yes, since connman's tethering supports only WPA2?

htot:
merged into https://github.com/htot/meta-intel-edison

PR #121, opened 2020-12-05
Zeus

htot:
This is a relatively big update where we upgrade Yocto from warrior to zeus and add btrfs support.
The documentation has been updated (a preview is [here](https://htot.github.io/meta-intel-edison/0.6-News.html) and will appear after merging [here](https://edison-fw.github.io/meta-intel-edison/0.6-News.html)).

Most notable is of course btrfs support, as it changes the emmc partition layout. Except for the rescue partition there is only one large partition. There are no fragmented unused partition pieces anymore. Instead we use snapshots for `@`, `@boot`, `@home`, and `@modules`. This allows to easily install and experiment with a 2nd kernel. It also allows a 2nd (alt) rootfs `@new`. This way we can reboot into an experimental rootfs (or potentially non-yocto, i.e. debian rootfs).

This has of course had a large influence on the way we flash the image, giving rise to the creation of a script `btrfsFlashOta.sh`, which allows installing image, kernel and U-Boot over the air.

### DIFF
--- a/meta-intel-edison-bsp/conf/machine/edison.conf
+++ b/meta-intel-edison-bsp/conf/machine/edison.conf
@@ -6,7 +6,7 @@
 # This sets compilation options close to what is used on android
 include conf/machine/include/tune-corei7.inc
 TUNE_CCARGS .= " -mstackrealign"
-DEFAULTTUNE = "corei7-32"
+#DEFAULTTUNE = "corei7-32"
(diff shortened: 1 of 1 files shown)


andy-shev (approved):

htot:
Thanks for the review @andy-shev , merged.

PR #126, opened 2021-03-15
Dunfell

htot:
This is the update to Yocto Dunfell. Documentation for now is here: https://htot.github.io/meta-intel-edison/

This time we have mostly fixes (btrfsFlashOta) and cleanups (removed clloader, sketch-check, mcu-fw).
Of course updates to U-Boot and Linux + added a preempt-rt kernel. Oobe is finally working again (point browser to http://edison/). Blink-led is now a kernel module.

And Debian image is now a btrfs snapshot, so you can dual boot Debian/Yocto.


andy-shev (approved):
I briefly looked into this and hadn't caught anything bad. Though couple of comments:
- perhaps you would need the revert patch for DWC3 ULPI (you were Cc'ed in the corresponding report)
- maybe we need to switch to LTS kernels and keep branches like devel (latest Yocto + Latest kernel) and per Yocto release with the corresponding LTS kernel.

htot:
Hi @andy-shev 
1. You mean "The fix e5f4ca3fce90 broke Intel Edison host mode" on linux-usb ML. Was there any follow up on that? I didn't consider that that might affect us. I will investigate.
2. That sounds like a very good idea. Do you think we should go back to 5.10? If you ask me 5.11.0 is working better then 5.10.0. But maybe 5.10.23 is even better.

andy-shev:
> 1. You mean "The fix e5f4ca3fce90 broke Intel Edison host mode" on linux-usb ML. Was there any follow up on that? I didn't consider that that might affect us. I will investigate.

Yes. It affects all OTG (automatically switched) boards. I have three different here :-)

> 2. That sounds like a very good idea. Do you think we should go back to 5.10? If you ask me 5.11.0 is working better then 5.10.0. But maybe 5.10.23 is even better.

For the release, I prefer to see v5.10.24+, yes. Leave newest kernel for devel/experimental branch.

htot:
> Yes. It affects all OTG (automatically switched) boards. I have three different here :-)

Ok, may be you can point me to the issue so I can reproduce by separate mail?

> For the release, I prefer to see v5.10.24+, yes. Leave newest kernel for devel/experimental branch.

But of course our images are a combination of kernel, U-boot and Yocto with each their own support policy. For instance Yocto Zeus is already no longer actively maintained. Yocto planned 2 year LTS for Dunfell but I didn't see this followed up, so might also EOL in April 2021.

So IMHO whatever we do an image will be frozen to a certain kernel release. But it does make sense to select a LTS kernel that has more or less calmed down.

andy-shev:
> > Yes. It affects all OTG (automatically switched) boards. I have three different here :-)
> 
> Ok, may be you can point me to the issue so I can reproduce by separate mail?

I dunno if you can reproduce on Arduino, but basically with that patch the host mode doesn't see the connected devices.
* Boot Edison in gadget mode
* Connect USB Mass Storage device (for example)
* No automatic recognition by a block layer -> no device node -> no nothing

> > For the release, I prefer to see v5.10.24+, yes. Leave newest kernel for devel/experimental branch.
> 
> But of course our images are a combination of kernel, U-boot and Yocto with each their own support policy. For instance Yocto Zeus is already no longer actively maintained. Yocto planned 2 year LTS for Dunfell but I didn't see this followed up, so might also EOL in April 2021.
> 
> So IMHO whatever we do an image will be frozen to a certain kernel release. But it does make sense to select a LTS kernel that has more or less calmed down.

You mean that Yocto itself chooses non-LTS kernel releases? Then we at least may do better. U-Boot OTOH doesn't make any difference (they simple don't have LTS).

htot:
> You mean that Yocto itself chooses non-LTS kernel releases? Then we at least may do better. U-Boot OTOH doesn't make any difference (they simple don't have LTS, effectivelly means whatever release of U-Boot on the release date of our stuff will be the best one, just don't update it in named branches).

Yocto Project provides for Dunfell linux 5.4.98. Gatesgarth adds 5.8.18. Indeed I think it is good do a more recent kernel in our layer.

But I meant Dunfell has a certain set of libs, systemd and other tools that after a time receive no more updates and are downloaded from certain sources. After a while it will not even build any more due to sources disappearing and people need to move to a newer Yocto release. That's why meta-intel-edison is constantly updated. And free of charge results to newer kernel :-)

The real value of a LTS kernel for us is that it has received lots of testing.

htot:
Merged

PR #133, opened 2021-09-04
remove sudo requirement from DFU operations

humanoid2050:
Things to note:
I am 90% sure I did the markdown links correctly. Perhaps you have a way to test those short of putting them online?
I left the invocations of `flashall.sh --recovery` with sudo as before. That uses the xFSTK tool, and I am not sure that the udev rule will help with that operation.


htot:
Thanks I will check before pushing. You can generate the site locally by using jekyl (it has a built-in web server)

humanoid2050:
Just ran it through Jekyll. Looks like the links work.

htot:
I merged this into https://github.com/htot/meta-intel-edison master and gatesgarth.
I hope to release that soon in which case it will appear here. Thanks!

PR #141, opened 2022-02-28
oobe_git.bb - configure_edison - re-introduce empty var dir

mwallnoefer:
Initially empty, but after the setup it gets filled by certain state files e.g. "password_setup.done".

### DIFF
--- a/meta-intel-edison-distro/recipes-support/oobe/oobe_git.bb
+++ b/meta-intel-edison-distro/recipes-support/oobe/oobe_git.bb
@@ -3,7 +3,7 @@ LICENSE = "MIT"
 
 SRC_URI = "git://github.com/edison-fw/edison-oobe;protocol=https"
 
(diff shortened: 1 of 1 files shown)


htot:
I merged into htot, that will soon be released. Thanks!

meta-intel-edison-distro/recipes-support/oobe/oobe_git.bb
```diff
@@ -14,6 +14,7 @@ RDEPENDS_${PN} = "python3-core python3-bottle"
 
 do_install() {
    install -d ${D}${libdir}/edison_config_tools
+   install -d ${D}/var/lib/edison_config_tools
```
htot, line 17: Even though this seems a bit strange, duplicated paths, hard coded.

mwallnoefer: I have just restored the old behaviour from Intel. `configure_edison` requires this path to be able to create the mentioned status files.

PR #142, opened 2022-03-07
Hardknott

htot:
This is to pull update from Gatesgarth to Hardknott. There are no dramatic changes except the obvious latest kernel, U-Boot and all the package updates from Hardknott (see 'docs: update manifests"). Other than that some smaller fixes, added CH341 driver and documentation updates.


andy-shev (commented):
Bot sure why revert is needed. Perhaps add a couple of words in the commit message?

andy-shev (commented):
Have we tried the latest available ACPICA to see if there are any issues?

andy-shev (commented):
Right, if any script uses hardware_id from U-Boot environment, they should be transformed to use fw_printenv or whatever the proper analogue for it.

andy-shev (commented):
Don't we have 5.15.xxx already available? Can we use stable release?

andy-shev (commented):
I would use the form of `0xhhhhhhhh_hexadecimal_latency_in_uS`

andy-shev (commented):
Don't we have some of these patches in one of the latest 5.15.xxx stable release? I believe we may switch to the latest and see what is left.

andy-shev (commented):
Can we use the latest U-Boot (as of today it should be v2022.04 I think)?

andy-shev (commented):
Didn't get why you need to duplicate installation of edison_config_tools folder. Does ${libdir} not work correctly? Where does it point to?

andy-shev (commented):
There are a few additional patches flying around to enable GPIO and I¹C on CH341  modules. Perhaps makes sense to add them as well.

andy-shev:
@htot  
Have commented the series. Sorry, it takes a bit longer.

htot (commented):
On Gatesgarth m4 failed to build, so I just copied in the version from Hardknott. That is actually ugly, and I should have written a .bbappend appending SRC_URI with the patch "file://0001-c-stack-stop-using-SIGSTKSZ.patch".

Anyhow, the reason to remove here is that we a building Hardknott, so the version from poky works and this duplicate can be removed. I will add to commit meassge.

meta-intel-edison-bsp/conf/machine/edison.conf
```diff
@@ -22,10 +22,7 @@ KBUILD_DEFCONFIG="i386_defconfig"
 KCONFIG_MODE="--alldefconfig"
 
 KERNEL_MODULE_AUTOLOAD += " ledtrig-heartbeat ledtrig-timer tun"
-# FIXME: file parameter should be based on partition UUID (from U-Boot) or fixed
-# with label (label seems to work, but driver is probed too early)
-#module_conf_g_multi = "options g_multi file=/dev/mmcblk0p9 stall=0 idVendor=0x8087 idProduct=0x0A9E iProduct=Edison iManufacturer=Intel"
-#KERNEL_MODULE_PROBECONF += "g_multi"
 
+PREFERRED_VERSION_acpica = "20210105"
```
htot, line 25: We have 2 sources: the poky layer and meta-acpi. meta-acpi in the hardknott branch and poky both have 20210105.
But it's a moving target. Poky Honnister has 20210730. So meta-intel-edison honnister (in my repo) is already on that one.

meta-intel-edison-bsp/recipes-bsp/u-boot/files/edison.env
```diff
@@ -16,7 +16,7 @@ ota_update_max_retries=3
 audio_codec_name=audio_codec="dummy"
 do_audio_support=setenv audio_support platform_mrfld_audio.${audio_codec_name}
 do_compute_target=if itest.b ${first_install_retry} -gt ${first_install_max_retries} || itest.b ${ota_update_retry} -gt ${ota_update_max_retries}; then echo "Switch to Rescue target"; setenv bootargs_target rescue; saveenv; fi
-mmc-bootargs=run do_bootargs_rootfs; run do_audio_support; setenv bootargs ${bootargs_common} ${bootargs_rootfs} ${bootargs_console} ${bootargs_debug} systemd.unit=${bootargs_target}.target hardware_id=${hardware_id}
```
htot, line 19: This env for U-Boot is supposed to provide hardwire_id to kernel command line. But afaikt always hardware_id=00. I am assuming that is wrong anyway and just dropped it. Correct?

meta-intel-edison-distro/conf/distro/poky-edison.conf
```diff
@@ -6,7 +6,7 @@ PREFERRED_PROVIDER_zlib ?= "zlib"
 PREFERRED_PROVIDER_zlib-native ?= "zlib-native"
 PREFERRED_PROVIDER_nativesdk-zlib ?= "nativesdk-zlib"
 # LTS kernel
-PREFERRED_VERSION_linux-yocto = "5.10.63%"
+PREFERRED_VERSION_linux-yocto = "5.15%"
 # test kernel
 #PREFERRED_VERSION_linux-yocto = "5.14%"
 # preempt_rt kernel
```
htot, line 12: In a later patch we provide for default kernel 5.15.25 LTS, for test kernel  5.16 and for preempt_rt 5.15.25-rt33. This is also a moving target. User can easily (and should) goto latest LTS by just renaming the file linux-yocto_5.15.25.bb to the latest. Of course for Honnister we will release with the latest available LTS at that time.

docs/_docs/Edison/Current State/4.2-networking.md
```diff
@@ -165,6 +165,14 @@ Run `cyclictest` on both Edison-Arduino's:
 ```
 cyclictest > /dev/null &
 ```
+`cyclictest` sets `/dev/cpu_dma_latency` to 0. You can not simply `echo` 0 to this file as it will return to its original value when `echo` terminate. However, the following will work:
+```
+exec 3<> /dev/cpu_dma_latency; echo "hex_latency_in_uS" >&3
```
htot, line 170: I can change that to 0xhhhhhhhh_hexadecimal_latency_in_uS

htot:
> Don't we have some of these patches in one of the latest 5.15.xxx stable release? I believe we may switch to the latest and see what is left.

After later linux-yocto patches in this series we have "linux-yocto: cleanup unused". I carefully checked that I didn't delete patches needed in any of 3 provided kernels.

htot:
> Can we use the latest U-Boot (as of today it should be v2022.04 I think)?

This is also a moving target, I would prefer to move to latest U-Boot in Honnister. That will also leave time to test it a bit.

htot:
> It's already v5.17 available. Perhaps switching to it is better than keeping v5.16.

Same, I like to have Honnister have latest LTS and track upstream latest release.

meta-intel-edison-bsp/recipes-kernel/linux/linux-yocto_5.15.25-rt33.bb
```diff
@@ -48,6 +48,7 @@ SRC_URI_append = " file://0041-enable-EXTCON_INTEL_MRFLD.cfg"
 
 # our additional configs
 SRC_URI_append = " file://ftdi_sio.cfg"
+SRC_URI_append = " file://ch341.cfg"
```
htot, line 51: I'd like to add "additional patches flying around to enable GPIO and I¹C on CH341 modules" but can't test myself. I don't mind to add if you tested or if at least they don't break the build. Which ones?

meta-intel-edison-distro/recipes-support/oobe/oobe_git.bb
```diff
@@ -14,6 +14,7 @@ RDEPENDS_${PN} = "python3-core python3-bottle"
 
 do_install() {
    install -d ${D}${libdir}/edison_config_tools
+   install -d ${D}/var/lib/edison_config_tools
```
htot, line 17: > Didn't get why you need to duplicate installation of edison_config_tools folder. Does ${libdir} not work correctly? Where does it point to?

`${libdir}` points `/usr/lib`, so `/usr/lib/edison_config_tools/public/` is where the web server template pages are installed. But for some reason /var/lib/edison_config_tools needs to exists otherwise the CLI is broken. python mysteries...

meta-intel-edison-bsp/recipes-bsp/u-boot/files/edison.env
andy-shev: Have you checked if any tool is actually using it for something more serious than debug printing?

meta-intel-edison-bsp/recipes-kernel/linux/linux-yocto_5.15.25-rt33.bb
andy-shev: I have no hardware to test, it was just a spontaneous idea. https://lore.kernel.org/all/20220401023306.79532-2-frank@zago.net/T/ Not sure if it’s the latest version of the series.

meta-intel-edison-bsp/recipes-bsp/u-boot/files/edison.env
htot: I see it is used in U-Boot: `setenv dfu_alt_info "ifwi${hardware_id} raw 0 8192 mmcpart 1;ifwib${hardware_id} raw 0 8192 mmcpart 2"`, but no idea why it was passed on the kernel command line. Something SFI related?

meta-intel-edison-distro/recipes-support/oobe/oobe_git.bb
mwallnoefer: Exactly, the _cli_ creates some temporary file there: https://github.com/edison-fw/edison-oobe/blob/07512dfcb97d77c57adb024c6ca3473abbae0229/src/configure_edison#L798

meta-intel-edison-bsp/recipes-bsp/u-boot/u-boot-common_2021.10.inc
```diff
@@ -27,6 +27,10 @@ SRC_URI += "file://btrfs.cfg"
 SRC_URI += "file://0001-REVERTME-usb-dwc3-gadget-skip-endpoints-ep-18-in-out.patch"
 SRC_URI += "file://0001-x86-tangier-Enable-support-for-SD-SDIO-family-in-the.patch"
 SRC_URI += "file://0002-x86-edison-Don-t-take-SD-card-detect-pin-into-consid.patch"
+SRC_URI += "file://netconsole.cfg"
```
htot, line 30: For some reason this patch is preventing `./ﬂashall.sh --recover` (and also manual attempts using `xfstk`) to work. I verified by reverting.

So in the final PR I will drop this, possibly until at a later time if we can figure out a way to have `netconsole` simultaneous with `xfstk` flashing.

docs/_docs/Edison/Introduction/0.6-News.md
```diff
@@ -5,12 +5,7 @@ sidebar: edison
...
- 3. Add ppp and openvpn support, see [Networking](4.2-networking.html#openvpn)
- 4. Fix Bitbake to allow signed feeds of DEB packages, see [Setting up for signed DEB repositories](5.0-Creating-a-deb-repository.html#setting-up-for-signed-deb-repositories)
- 5. Create recipe for `zerotier`
- 6. Restore SD card access in U-Boot and Linux
- 7. Finally fix HSU for communication speeds up to 3.5Mb/s, see [HSU](4.4-hsu-i2c-spi.html#hsu)
- 8. Update Debian snapshot to bullseye, see [Debian](2.6-Building-Debian)
+ 1. Linux make 5.15.25 LTS the default kernel, upgrade U-Boot to v2021.10
+    We also provide v5.16.0 (testing) and v5.15.25-rt33 (LTS PREEMPT_RT kernel) for development purposes.
+ 2. Update to Yocto Hardknott.
+ 3. Move `xfstk` recipe to meta-intel-edison-distro to make it visible on xfstk: move to meta-intel-edison-distro
```
htot, line 11: This is obviously a type, should say to make it visible on [https://layers.openembedded.org](https://layers.openembedded.org/layerindex/branch/hardknott/recipes/)

meta-intel-edison-bsp/recipes-kernel/linux/linux-yocto_5.15.25-rt33.bb
htot: I see this is brand new v5 = 1-apr-2022. I could add to Honnister or it will go into a new kernel anyway and maybe Marco V (on Telegram) can test?

meta-intel-edison-bsp/recipes-bsp/u-boot/files/edison.env
andy-shev: No idea, that's why I asked to check the user space tools in the original image to see if any of them is using it (it needs to be checked in source code, because some C code might parse it).

PR #148, opened 2022-10-29
Minor fixes to the documentation

macvenez:
Changed btrfs-tools package to btrfs-progs because it seems to have changed name (btrfs-tools no more exists).
Added "cd my_Edison_Workspace" otherwise you were creating the build directory without using it and everything would have compiled in your home dir


htot:
I've pulled this into my htot repo, soon to be released here. Thanks.

PR #154, opened 2023-11-26
Update oobe_git.bb

dcacklam:
Since Sumo32 is frozen in time & the current oobe source doesn't work with it, have this recipe use the last version that it knows how to build correctly (and that doesn't have post-Sumo dependencies - eg python3) 

This, along with a change to the URL for the ACPI support package, allows it to build again.

Doing this because there are still some things 'out there' that use MRAA & the Edison - most of them running on 3.x Linux kernels... Making sumo32 build again makes it easier to use it's kernel image.


htot:
Nice to here that there are people out there using the EOL version and providing some CPR. I didn't test this but LGTM, thanks!

I have been hoping people would be providing an MRAA layer on top of IIO drivers in the kernel to provide some sort of compatibility layer, but it seems development has slowed there. Alas.

PR #155, opened 2023-11-27
Update setup.sh

dcacklam:
Fix broken patch that prevents sumo32 from building. Patch has no discernable impact on functionality when removed.


htot:
LGTM, thanks!

PR #160, opened 2025-01-27
Scarthgap

htot:
This updates from Yocto Kirkstone -> Langdale -> Mickledore -> Nanbield -> Scarthgap which is a LTS .

What's new is [here](https://htot.github.io/meta-intel-edison/0.6-News.html).


meta-intel-edison-distro/recipes-devtools/ninja/files/ninja.py
```diff
@@ -21,6 +21,7 @@
                 NewArgv.append(i)
 
     if seen:
+        w = os.open(fifoname, os.O_WRONLY|os.O_NONBLOCK)
```
andy-shev, line 24: But is this `w` used anywhere? You can run `pylint` to see the warning it issues on this line (of course it may be more on the rest).

utils/0001-u-boot-Fix-path-to-merge_config.sh.patch
```diff
@@ -1,9 +1,9 @@
-From afa3f6029bb6c232cde8cae80073b8ee0749b0bb Mon Sep 17 00:00:00 2001
+From 338b578e4c24860d7adf94ac04df4479c1acd52f Mon Sep 17 00:00:00 2001
 From: Ferry Toth <ftoth@exalondelft.nl>
-Date: Wed, 21 Dec 2022 22:39:23 +0100
-Subject: [PATCH 1/1] [PATCH] u-boot: Fix path to merge_config.sh
+Date: Sun, 22 Dec 2024 20:02:46 +0100
+Subject: [PATCH 1/1] u-boot: Fix path to merge_config.sh
 
-Current u-boot recipe in kirkstone fails during do_configure because
+Current u-boot recipe in scarthgap fails during do_configure because
```
andy-shev, line 6: Hmm... Shouldn't be this fixed in the Yocto Poky? Or did I get this incorrectly?

meta-intel-edison-bsp/recipes-support/gadget/files/conf-gadget.sh
```diff
@@ -74,5 +74,5 @@ echo 120 > configs/c.1/MaxPower
 
 # Activate gadgets
 echo dwc3.0.auto > UDC
-sleep 1
```
andy-shev, line 77: Basically you need to watch for a file in sysfs to appear (but which file, I don't know, perhaps the one that gives a new interface under net?).

```
    # Wait until the interface has come up
    for pass in $(seq 3); do
        for eth in $(find /sys/class/net -type l -name 'eth*' -exec basename {} \;); do
```

setup.sh
```diff
@@ -259,12 +259,6 @@ COPYLEFT_LICENSE_INCLUDE = 'GPL* LGPL*'
   fi
 
   case $my_sdk_host in
-    win32)
-      extra_conf="SDKMACHINE = \"i686-mingw32\""
-      ;;
-    win64)
```
andy-shev, line 265: Does anybody try to build it under WSL2? I am just curious...

meta-intel-edison-bsp/recipes-kernel/bcm43340/bcm43340-fw.bb
```diff
@@ -21,15 +23,15 @@ LIC_FILES_CHKSUM = "file://edison-firmware/broadcom_cws/wlan/firmware/LICENCE.br
...
-FILES:${PN} += "/lib/firmware/brcm/*"
+FILES:${PN} += "${libdir}/firmware/brcm/*"
 
 inherit allarch update-alternatives
 
 do_install() {
-        install -v -d  ${D}/lib/firmware/brcm/
-        install -m 0755 edison-firmware/broadcom_cws/wlan/firmware/bcmdhd_aob.cal_4334x_b0 ${D}/lib/firmware/brcm/brcmfmac43340-sdio-fr.txt
-        install -m 0755 edison-firmware/broadcom_cws/wlan/firmware/bcmdhd.cal_4334x_b0 '${D}/lib/firmware/brcm/brcmfmac43340-sdio.Intel Corporation-Merrifield.txt'
-        install -m 0755 linux-firmware/cypress/cyfmac43340-sdio.bin ${D}/lib/firmware/brcm/brcmfmac43340-sdio.bin
```
andy-shev, line 32: The Cypress FW are available in linux-firmware repository, I'm wondering why we don't use them directly.

andy-shev (changes requested):
I have done a brief review, comments here and there and some can be addressed.

meta-intel-edison-distro/recipes-devtools/ninja/files/ninja.py
htot: Not by our process, but the child (make/ninja) needs it open non-blocking and childs are opening and closing this. I open this one just to make sure the pipe remains. I don't know if that's a bug in our ninja fork (I think upstream they are still struggling to get the jobserver aligned with make). But without this patch it doesn't work.

Note: the whole jobserver stuff is needed as normally you have make orchestrating submakes. We have bitbake orchestrating submakes and subninjas. With this enabled we never have more gcc's running than threads. Now linux has no problem scheduling more gcc's than threads, but each consumes memory, especially g++ like 1GB. So on my machine 16 HT leads to 4 x 16 x 1GB. I used to have 16GB (now 32GB)...

meta-intel-edison-distro/conf/distro/poky-edison.conf
```diff
@@ -17,15 +17,14 @@ PREFERRED_VERSION_linux-yocto = "6.1.55%"
...
-DISTRO_FEATURES = "systemd pulseaudio alsa argp bluetooth ext2 largefile usbgadget usbhost wifi xattr zeroconf pci pam ppp"
+DISTRO_FEATURES = "systemd pulseaudio alsa argp bluetooth ext2 largefile usbgadget usbhost wifi xattr zeroconf pci pam ppp usrmerge"
 
 # Disable sysvinit for recipes with systemd support
 DISTRO_FEATURES_BACKFILL_CONSIDERED += "sysvinit"
 VIRTUAL-RUNTIME_initscripts = ""
 VIRTUAL-RUNTIME_init_manager = "systemd"
 
 # This builds tools and python3 bindings for libgpiod
-PREFERRED_VERSION_libgpiod = "2.0"
```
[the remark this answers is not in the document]
htot: Indeed. When I found upstream had updated libgpiod I dropped the local version. Then while fixing the usrmerge thing I found the PREFERRED_VERSION superfluous (the highest revision is automatically preferred). So, that was more like a "while here" clean up. No functional change intended.

setup.sh
```diff
@@ -320,7 +320,8 @@ COPYLEFT_LICENSE_INCLUDE = 'GPL* LGPL*'
   echo "Cloning meta-qt5 layer to ${oe_dir} directory from local cache"
   git clone ${my_dl_dir}/meta-qt5-mirror.git meta-qt5
   cd ${oe_dir}
-  git checkout ${yocto_tag}
+#  git checkout ${yocto_tag}
```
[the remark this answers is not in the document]
htot: Yes introduced in "meta-qt5: [revertme] checkout master while nanbield awaits release"

utils/flash/btrfsSnapshot.sh
```diff
@@ -7,7 +7,7 @@
 
 dirs="bin boot dev etc home lib media mnt opt proc run sbin sketch sys tmp usr var"
 deploy_dir=tmp/deploy/images/edison
-image_name=edison-image-edison
+image_name=edison-image-edison.rootfs
```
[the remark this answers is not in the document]
htot: The Yocto wizards changed the file name of the image. Comment would be useful, true.

meta-intel-edison-bsp/recipes-kernel/bcm43340/files/bluetooth_bd_addr.service
```diff
@@ -1,13 +1,20 @@
...
-ExecStart=/usr/bin/bluetooth_bd_addr.sh
+Type=oneshot
+RemainAfterExit=true
+TimeoutStartSec=60
 Restart=on-failure
+ExecStart=/usr/bin/bluetooth_bd_addr.sh
 StandardOutput=journal+console
+StandardError=journal
+# Redirect stdin to ensure btmgmt doesn't hang
+StandardInput=tty
```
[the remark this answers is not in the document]
htot: Seems they change systemd at some point so there is no stdin / stdout by default. Then they spawn a child process for bt tools, that actually don't work without stdin. This options changes back to having the needed stdin. Why the bt tools need a stdin beats me. Complicated code.

meta-intel-edison-distro/recipes-support/battery-voltage/files/battery-voltage.c
```diff
@@ -2,16 +2,15 @@
...
 #include <stdlib.h>
 #include <stdio.h>
 #include <errno.h>
 #include <unistd.h>
 #include <sys/stat.h>
 #include <fcntl.h>
-
-#define uchar unsigned char
-#define uint  unsigned int
+#include <iio.h>
```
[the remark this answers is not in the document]
htot: It's even worse. I included stdio and stdlib while they were already included above :-(

meta-intel-edison-bsp/recipes-kernel/bcm43340/files/bluetooth_bd_addr.service
```diff
@@ -10,8 +10,8 @@ RemainAfterExit=true
 TimeoutStartSec=60
 Restart=on-failure
 ExecStart=/usr/bin/bluetooth_bd_addr.sh
-StandardOutput=journal+console
-StandardError=journal
+StandardOutput=journal
```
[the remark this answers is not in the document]
htot: Yes should have been squashed with previous commit with the same name. The logging to console was for debugging.

meta-intel-edison-bsp/recipes-kernel/linux/files/0001a-8250_mid-toggle-IO7-on-ttyS1-interrupt-entry.patch
```diff
@@ -0,0 +1,56 @@
...
+
+diff --git a/drivers/tty/serial/8250/8250_mid.c b/drivers/tty/serial/8250/8250_mid.c
+index 8ec03863606e..0be5ddfe8051 100644
+--- a/drivers/tty/serial/8250/8250_mid.c
++++ b/drivers/tty/serial/8250/8250_mid.c
+@@ -7,6 +7,7 @@
+  */
+ 
+ #include <linux/bitops.h>
++#include <linux/gpio.h>
```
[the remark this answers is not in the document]
htot: I'll have a look. This patch BTW was just toggling a pin in the kernel for my serial port interrupt measurements.

setup.sh
htot: It should be possible https://docs.yoctoproject.org/5.0/singleindex.html#setting-up-to-use-windows-subsystem-for-linux-wsl-2

But with headaches...

meta-intel-edison-bsp/recipes-kernel/bcm43340/bcm43340-fw.bb
htot: That is what we do here, line #14 points to the repo which causes that to be cloned and checked out.
Then we override the do_install() to install cyfmac43340-sdio.bin  to the install dir. We rename it brcmfmac43340-sdio.Intel Corporation-Merrifield.bin what the kernel seems to be looking for (is that wrong?).
From there it gets packaged automatically.

meta-intel-edison-distro/conf/distro/poky-edison.conf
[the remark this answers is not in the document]
htot: Fixed in Nanbield-v2

meta-intel-edison-distro/recipes-support/cleanjournal/files/clean_journal.sh
```diff
@@ -37,7 +37,7 @@ max_allowed_free_space=10
 
 # check remaining space and update clean_journal_needed variable
 check_free_space() {
-    current_free_space=$(df -h | grep /dev/root | awk '{print 100 - $5}' | sed 's/%//')
+    current_free_space=$(df -h | grep /$ | awk '{print 100 - $5}' | sed 's/%//')
```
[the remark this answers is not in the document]
htot: Ok in scarthgap-v2

utils/0001-u-boot-Fix-path-to-merge_config.sh.patch
htot: This is a refresh of an old patch we have since 2019. In principle you are right. However, we have our own U-Boot recipe (based on the old Intel Edison one). And Yocto has something newer. In our case configure fails because we do include their newer configure code. I guess they don't have this problem, so we must be doing something else wrong.
Likely the real solution would be to use the upstream recipe to build U-Boot, if needed to use all the Yocto trickery to get the Edison version the same as we have now.
But for now, I'd like to leave this as is.

meta-intel-edison-bsp/recipes-support/gadget/files/conf-gadget.sh
htot: `for usb in $(find /sys/class/net -type l -name 'usb*' -exec basename {} \;); do`?

andy-shev: I think even for USB it still will be `eth` in the path.

htot: I think it should be like this:
```
# Wait until the interface has come up
    for pass in $(seq 5); do
	readlink -q -e /sys/class/net/usb*
	if [[ $? -eq 0 ]]; then
            break
        fi
        sleep 1; 
    done
```
Check if the symbolic link exits and point to a file.

andy-shev: > I think it should be like this:
> 
> ```
> # Wait until the interface has come up
>     for pass in $(seq 5); do
> 	readlink -q -e /sys/class/net/usb*
> 	if [[ $? -eq 0 ]]; then
>             break
>         fi
>         sleep 1; 
>     done
> ```
> 
> Check if the symbolic link exits and point to a file.

You don't need that `[[ ]]` thingy, just `if readlink ...; then`.

PR #163, opened 2025-05-14
Fix serial port login in Ubuntu

lybtongji:
fix serial port login


htot:
Hi, I am completely replacing generation of Debian / Ubuntu by enabling lxc/docker to install Debian / Ubuntu containers. Therefor this manual page is entirely dropped, and the patch no longer needed. Thanks.

PR #167, opened 2025-12-18
Correct Ubuntu version for Yocto Scarthgap build

Divyesh-Kamalanaban:
Had an incorrect ubuntu version (24.10 instead of 24.04) for Ubuntu Noble. Changed it in the markdown file.


htot:
Thanks, will appear soon.

PR #169, opened 2026-04-05
Midlife update scartgap

htot:
This is a midlife update for scarthgap. So there are some fixes in our recipies as well as updates for the kernel. Scartgap itself will be the latest version when you check this out.

There are also some dropped code (generating debian images) and new features (`lxc`, `docker` and `snap` support). As well as a big update of the documentation.

See the overview here: https://htot.github.io/meta-intel-edison/0.6-News.html.

Tested both x86_64 and i686 builds (the latter tested except `docker` and `lxc` as there are no i686 images available).


meta-intel-edison-bsp/recipes-kernel/linux/linux-yocto_current.bb
andy-shev, line 73: Shouldn't be needed, I upstreamed that workaround

andy-shev (commented):
I briefly looked parts related to kernel and found nothing problematic.

